Issue #9, opened 2018-09-25
course calendar lookup function


jidicula:
# Bug Report: Full-year course lookup

### Bug Description
Can't look up courses with the form DEPT 100D1

### Reproduction Steps
1. Go to any channel.
2. Send `?course anat 690d1`
3. Marty responds with: `No course found for anat 690d1.`

### Potential Reasons behind Bug:
Marty doesn't accept course numbers that have non digits in them

PR #10, opened 2018-09-26
[WIP] Edit regex pattern for ?course command

trungams:
Fix to issue described in #9


trungams:
why is the check failing? D:

PR #14, opened 2018-10-09
fixed #13

### LINKED ISSUE #13: Urban HTML decoding
# Bug Report: urban HTML decoding

### Bug Description
Not decoding HTML escape sequences.

### Reproduction Steps
*Steps to reproduce the behavior, for example:*
1. Go to any channel.
2. Send ?urban [any query which has a definition containing an HTML escape sequence]
3. Marty responds with a definition including &amp; &apos; etc

### Potential Reasons behind Bug:
HTML escape sequences aren't decoded. Probably can fix with some BS4 tweaking but tbh I'm just trying to get a tshirt for hacktoberfest.

jidicula:
ugly hack to replace amperand and apostrophe escape sequences that BS4
doesn't catch for some reason.

### DIFF
--- a/cogs/helpers.py
+++ b/cogs/helpers.py
@@ -207,6 +207,8 @@ class Helpers():
             return
         word = word.get_text()
         definition = soup.find('div', {'class': 'meaning'}).get_text()
+        definition = definition.replace("&amp;", "&")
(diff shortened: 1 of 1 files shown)


trungams:
Closing this PR. I changed ?urban command in #19

PR #101, opened 2019-05-07
Update pipfile and .travis.yml

trungams:
- specify version of dependencies in Pipfile
- ignore existing virtual environment during travis build


trungams:
yeah I wasn't aware until now either. Check my comments in #94 for details

PR #103, opened 2019-05-08
add some documents and minor codestyle change

trungams:
- Create contributing.md and pull_request_template.md
- Remove duplicate imports of `asyncio` in Main.py
- Move log file pointer to instance attribute of memes object

### DIFF
--- a/cogs/memes.py
+++ b/cogs/memes.py
@@ -11,6 +11,10 @@ import random
 class Memes:
     def __init__(self, bot):
         self.bot = bot
+        self.mix_command_logger = open('mix.log', 'a')
+
+    async def on_disconnect(self):
+        self.mix_command_logger.close()
 
     @commands.command()
     async def lenny(self, ctx):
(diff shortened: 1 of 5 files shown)


cogs/memes.py
```diff
@@ -11,6 +11,10 @@
 class Memes:
     def __init__(self, bot):
         self.bot = bot
+        self.mix_command_logger = open('mix.log', 'a')
+
+    async def on_disconnect(self):
+        self.mix_command_logger.close()
```
idodin: If the python process running the script crashes for some reason, there's no guarantee that it'll be run. I think discord-py will run it before exiting in the case of a runtime error though...

trungams: @davidlougheed it's here: https://discordpy.readthedocs.io/en/rewrite/api.html?highlight=on_disconnect#discord.on_disconnect

IIRC the discord.py client has a handler for bot closing event

davidlougheed: I can't find info on this in the docs; do you know if this is guaranteed to be called even if there's a crash?

davidlougheed: ok, we may want to manually flush the log occasionally then, but not a big deal.

idodin: Just as a note, IIRC from my other bot, discord completes disconnect on runtime error so the on_disconnect should still be called.

Issue #81, opened 2019-02-10
Add a poll feature/Figure out how to do it


le-potate:
### Feature Description:
Several members including the **_server owner_** have asked for an anonymous poll feature. However, we aren't quite sure how to implement it

### Implementation Suggestion:
- Since the classic Discord way to have polls is through emojis, the "familiar" way would be to count reacts and reset upon click. However, the time to make the reacts reset would be a problem if the poll is to be truly anonymous since people could see it during that time
- A better way would be to answer a DM from the bot. However, this might be a more annoying way.
- There could be multiple "versions", like a non-anonymous one that just removes emojis that aren't choices, etc.

trungams:
<@!269208707068329994> ping pong

le-potate:
<@!259549018781253632> pong ping

le-potate:
Seems like discord will do it's own version/give more appropriate tools in the near future: see https://blog.discord.com/the-future-of-bots-on-discord-4e6e050ab52e and this demo: https://www.figma.com/proto/DDsdQLvBqmmPgZNgeNKmvf/Bot-Proposal?node-id=129%3A77&viewport=28579%2C-4007%2C6.748552322387695&scaling=min-zoom

I think for now this is wontfix, but this could be opened again if we get rid of Mose before those features are implemented by Discord

PR #106, opened 2019-05-12
WIP Adding Polls

le-potate:
Adding a poll command as mentioned in #81 
There is no anonymous function yet: it only serves to format a poll, dividing the options into pages and adding reactions to the message so that nobody has to lose his vote.
We need to figure out if the while loop at the end is acceptable. I think it is as it will not actually iterate through it all the time continuously: it only does a loop when someone reacts in the channel, the rest of the time it's on this line https://github.com/idoneam/Canary/blob/e9c47910fc61c8c7cdf7e35d7f47624e28b1c1ab/cogs/poll.py#L300
Right now I put a limit of 1 day for the maximum duration but I think it could be even longer if it turns out that, indeed, this while loop doesn't take a lot of ressources.


cogs/poll.py
```diff
@@ -0,0 +1,318 @@
...
+        p = Pages(ctx, msg, options_list, embed, timeout, timeout_unit, failed)
+
+        await p.paginate()
+        return
+        # need to also eventually add options for anonymous voting in DMs
+        # restricting to only one choice only and
+        # add a duration for polls.
+
+
+class Pages:
```
trungams: If this class behaves similarly to Pages inside `paginator.py`, better put them in the same file and give them different names or make one class inherit from the other, or modify the original class to allow longer time out. This way you don't need to write much more code, and other cogs can benefit from it as well

cogs/poll.py
```diff
@@ -0,0 +1,318 @@
...
+        self.message = msg
+        self.options_list = options_list
+        self.embed = embed
+        self.failed = failed
+        self.actions = [('◀', self._prev_page), ('▶', self._next_page)]
+        self.currentPage = 0
+        self.lastPage = len(options_list) - 1
+        self.timeout = timeout
+        self.timeout_unit = timeout_unit
+        self.timeout_end = time.time() + timeout * TIME_UNITS[timeout_unit]
```
trungams: In almost all cases, the original paginator is used for displaying static content, so it's negligible if the bot fails while waiting to paginate. However, keeping longer timeouts means you need a better solution to deal with the bot shutting down randomly, and it's not really a good design to wait for, say, a day just to update a message. You don't have to handle all cases which rarely happen in reality. Is there a better way to handle timeout?

cogs/poll.py
```diff
@@ -0,0 +1,318 @@
...
+    def __init__(self, bot):
+        self.bot = bot
+
+    @commands.command()
+    async def poll(self, ctx, *args):
+        """
+        Creates a poll with a maximum of 18 options.
+        ==Examples==
+        ?poll "Question" "Option 1" "Option 2" "..."
+        ?poll -t 30 minutes "Question" "Option 1" "..."
```
trungams: Instead of having "question" and -t which denotes the beginning of different input types, try to make them more consistent ("question" could be changed to -q for example), and I think there should be a better way to parse arguments

cogs/poll.py
```diff
@@ -0,0 +1,318 @@
...
+import time
+
+ALPHABET = list(string.ascii_uppercase)
+EMOJI_ALPHABET = [
+    u"\U0001F1E6", u"\U0001F1E7", u"\U0001F1E8", u"\U0001F1E9", u"\U0001F1EA",
+    u"\U0001F1EB", u"\U0001F1EC", u"\U0001F1ED", u"\U0001F1EE", u"\U0001F1EF",
+    u"\U0001F1F0", u"\U0001F1F1", u"\U0001F1F2", u"\U0001F1F3", u"\U0001F1F4",
+    u"\U0001F1F5", u"\U0001F1F6", u"\U0001F1F7"
+]    # :regional_indicator_a: to :regional_indicator_r:
+TIME_UNITS = {
```
trungams: Personally I don't like having different time units. What if you only accept minutes? A poll means nothing if it last only 10 seconds, and it's quite easy to convert hours -> minutes. It also makes things way easier to parse

cogs/poll.py
```diff
@@ -0,0 +1,318 @@
...
+            Default: 1 day
+        """
+        question = ""
+        # default timeout:
+        timeout = 1
+        timeout_unit = "day"
+        skip = 0
+        choices = []
+        # look for switches
+        for arg in range(len(args)):
```
trungams: minor styling, change `arg` to `index`, `idx` or simply `i`

cogs/poll.py
le-potate: The problem is there's a couple major differences. An important one is that this one takes a message that was already sent and edits it. A main reason for that is that the reacts for the poll options must already be placed before the paginator starts. I am not sure this would be useful for other stuff than polling. Another significant difference is that it takes as input a 2D array where the items are already divided: the rows are the different pages and the columns are the items on each page (The original paginator divides the items itself). Finally, unlike the original paginator, it only edits the fields, so that the embed attributes (author, etc) aren't changed.

le-potate: If I shouldn't let it in poll.py, then I think the best option is for me to make another class in paginator.py considering all these differences, but I'm gonna need to generalize it before I can do that, because right now there's a lot of poll-specific things in it. Also, if I do that, there's things I won't be able to ever implement, like tracking what the current option with the most votes is, because it would need to be in paginator

cogs/poll.py
le-potate: I'm not sure I understand, question doesn't denote the beginning of an input type. The way it works is that the first argument that is not a switch, is the question, like this:
?poll "What do you think of that?" "Yes it's great" "No it's bad" 
And anywhere in this you could write -t 30 minutes

cogs/poll.py
le-potate: I agree about that a polls in seconds isn't useful, I just put it there because, why not since I had already put all the others,  but I'm not sure converting hours into minutes is something the user should have to think about.. Like if the poll timeouts after a day, it's way better to write "1 day" than "1440 minutes"

cogs/poll.py
le-potate: The reason it takes a boolean "failed" as input is that since the poll puts a lot of emojis on the message, it's really easy to get to a point where users only need to put a couple emojis and the bot won't be able to paginate. And I don't think it's a rare thing because people on discord do like to put emojis on messages 😄 Before calling paginate, poll does try to take care of that by clearing emojis and trying again. However, if it isn't unable too, I still want the poll to print, which is why paginator gets called with failed=True. Paginator will then just print every options on the same page.

cogs/poll.py
trungams: ok yeah I misunderstood this part

le-potate: I will still make the examples clearer in case

cogs/poll.py
trungams: Counting votes is simple as long as you have the original message id. Adding emojis for options before adding the arrows can be implemented in the original paginator.

cogs/poll.py
trungams: That's not what I was talking about. I was referring to situations where the bot is shut down while a poll is active on the server

le-potate: Oh sorry, I see what you meant now. So I'd need to keep the message IDs of the polls in the database if I want to be able to make pages work again after the bot restarts?

le-potate: If yes this is is a good idea, because I will need to use the database anyway when I write anonymous polling by DM.

trungams: Yes. Save the message id in the database and store it in memory on startup. I'm also suggesting that public voting should be done via a command so essentially the bot will not have to loop until poll timeout like it's doing now for paginating.

le-potate: Agreed about public voting via a command, but I still think that I should keep a simple version with emojis without paginator, maybe with a switch like "-s" for simple. Paginator anyway was already not that essential because people won't write polls that are very long, and even if they did the maximum is still less than 2000 characters, but I could even put a lower maximum for that mode

le-potate: Agreed about public voting via a command, but I still think that I should keep a simple version with emojis without paginator, maybe with a switch like "-s" for simple. Paginator anyway was already not that essential because people won't write polls that are very long, and even if they did the maximum is still less than 2000 characters, but I could even put a lower maximum for that mode

le-potate: I think it's nice to have something that just puts the emojis for you so you don't lose your vote but it's still simple and don't require people to use commands.

cogs/poll.py
le-potate: I've decided that modifying paginator.py was the best way, adding a display_option that takes a list of embeds, an option to replace the message by a string or an embed at the end instead of deleting, and a way to input a list of emojis to put before or after the arrows. I'll commit it soon, but for now I've tested it on my machine and it works well.

cogs/poll.py
trungams: Yeah. That would be great

cogs/poll.py
trungams: Doesn't an option for embed formatting already exist? You can modify it without needing to add new attributes. I don't think it's going to contrast any user of paginator

le-potate:
Since a lot changed in the dev branch and I learned stuff about database/paginator, I prefer to restart from scratch

Issue #108, opened 2019-05-19
[Enhancement] Generalize bot


le-potate:
Since we tell people how to use the bot on their server, a few things should be different
-Find all the direct mentions of server/channel IDs, upmartlet/downmartlet, roles (idoneam and Discord Moderator are the only roles referred to afaik), and take those values from config instead
-Maybe do the same thing for all mentions of "marty" (except variables, name of the db, etc), take it from a botName in the config file instead
-In main.py, take "command_prefix" from config
-Add something to the readme to say to change those values in the config file
There are not that many files with things like that, so far I've found:
main.py, mod.py, reminder.py, score.py
but if you see anything else please tell me in the comments

le-potate:
I think the only one of those I won't be able to do is command_prefix because it needs to be defined before the parser is called

trungams:
Define a class Canary (or however you want to call it) that inherits discord.py's bot object. Do all the parsing inside `__init__()`, then you can do whatever you want

trungams:
that should be a better way to implement a custom discord bot, we're just lazy

le-potate:
Right now it only calls parser.Parser() if you run the file, not if you import it, but if I do that it will parse everytime right? Is that okay?

idodin:
This should be broken down into smaller tasks though - in order to better keep track on who is doing what :p

And yeah, when I wrote the Parser, I knew that I should probably override __init__, I was just lazy and excited to get done with config hehe.

idodin:
Whoops, I actually misread "Participants" for "Assignees" @le-potate - its fine if you want to tackle this on your own; but break down into smaller Issues in any case so that we can see what's going on more easily in commits :) .

idodin:
> Right now it only calls parser.Parser() if you run the file, not if you import it, but if I do that it will parse everytime right? Is that okay?

`parser.Parser()` is a Constructor. If you use it inside the `__init__()` function of a new "specific Bot" Class, it will only be run whenever you construct the Bot object - which is the desired behaviour :o

le-potate:
No I agree, I didn't think about the stuff in main.py that happens before it's parsed so I thought this would be something I could do alone, I'm nearly finished with the other files, so i'm gonna commit that and then you guys can figure out main.py

le-potate:
@idodin @trungams Branch issue-108 now contains the modified config.ini, parser.py, mod.py, reminder.py and score.py file. There's only Main.py left

trungams:
[...]

PR #109, opened 2019-05-27
Generalize bot

le-potate:
As discussed in issue #108

### DIFF
--- a/cogs/mod.py
+++ b/cogs/mod.py
@@ -29,8 +29,11 @@ class Mod():
     @commands.command()
     async def answer(self, ctx, *args):
         if isinstance(ctx.message.channel, discord.DMChannel):
-            message = "{}".format(" ".join(args))  # to work regardless of whether the person uses apostrophes
-            channel_to_send = self.bot.get_channel(self.bot.config.reception_channel_id)
+            message = "{}".format(
+                " ".join(args)
+            )    # to work regardless of whether the person uses apostrophes
+            channel_to_send = self.bot.get_channel(
+                self.bot.config.reception_channel_id)
             msg = '{} 📣 {}'.format(str(ctx.author.name), message)
             await channel_to_send.send(content=msg)
             await ctx.send("`Message sent`")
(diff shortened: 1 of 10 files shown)


config/config.ini
```diff
@@ -20,7 +20,23 @@
 [Discord]
 Key = KEY
 
+[Server]
```
trungams: remove the names and IDs specific to the discord server. if you want to generalize things you need to be consistent. and wasn't this file supposed to be ignored?

cogs/mod.py
```diff
@@ -21,32 +21,38 @@
...
-            return
-        ctx = await self.bot.get_context(message)
-        if ctx.command:
-            return
-        if isinstance(message.channel, discord.DMChannel):
-            channel_to_send = self.bot.get_channel(454061583874785280)
-            msg = '{} 📣 {}'.format(str(message.author), message.content)
-            await channel_to_send.send(msg)
+    @commands.command()
+    async def answer(self, ctx, *args):
```
trungams: I need a mod's say on this change @mjip

config/config.ini
davidlougheed: It’s still generalized if specific stuff is relegated to config and can be changed easily. Be careful about removing since then the auto deploy will break unless you guys have implemented a fix for that

davidlougheed: (to clarify: i think it's best if this stuff isn't hardcoded in config, although it doesn't matter too much, but you'll have to make sure the deployment system is good [and also generalizable lol] first)

le-potate: Yeah, I forgot but I wanted to change the readme too, right now it says go to config.ini to change the token but now it should say that there are other values to change. And yeah this is already better than what we had before, where you had to go to each file individually to and change every direct mention of the server ID, roles etc, but if you guys have a better way go for it

cogs/mod.py
le-potate: (This was discussed in #idoneam with @jidicula)

PR #125, opened 2019-08-15
Issue 122

jidicula:
New feature: CFIA RSS feed subscription

## Description
Bot will post food recall warnings as they're issued.

## Motivation and Context
Eating rotten food sucks. It could also kill you.

## How Has This Been Tested?
Ran on my local machine using a test token. Behaviour was as expected.

## Types of changes

## Checklist:


cogs/helpers.py
```diff
@@ -38,18 +36,61 @@
...
+        """
+        await self.bot.wait_until_ready()
+        while not self.bot.is_closed():
+            food_channel = utils.get(
+                self.bot.get_guild(self.bot.config.server_id).text_channels,
+                name="food")
+            newest_recall = feedparser.parse(CFIA_FEED_URL)['entries'][0]
+            newest_recall_id = newest_recall['id']
+            try:
+                id_unpickle = open("pickles/recall_tag.obj", 'rb')
```
davidlougheed: will I have to create the pickles directory? if so, can there be a check to create the pickles directory if needed?

cogs/helpers.py
```diff
@@ -38,18 +36,61 @@
...
+        # Written by @jidicula
+        """
+        Co-routine that periodically checks the CFIA Health Hazard Alerts RSS
+         feed for updates.
+        """
+        await self.bot.wait_until_ready()
+        while not self.bot.is_closed():
+            food_channel = utils.get(
+                self.bot.get_guild(self.bot.config.server_id).text_channels,
+                name="food")
```
davidlougheed: Maybe this should be a configurable thing so that it isn't hardcoded to a specifically-named channel. Fine for now, but maybe an issue should be created for a future pr

cogs/helpers.py
jidicula: I created the pickles directory and pushed the object actually. This isn't a great longterm solution – I can add an os.mkdir with a try/except.

cogs/helpers.py
jidicula: Yeah, it should definitely go in config.ini. I'll make another PR for both this and the above suggestion for pickles.

cogs/helpers.py
davidlougheed: sure, that would be good. and probably add pickles/* to gitignore

cogs/mod.py
```diff
@@ -48,8 +48,8 @@ def __init__(self, bot):
         """
         dest = user
         await dest.send(
-            content='{}\n*To answer write* `{}answer "your message here"`'
-            .format(message, self.bot.config.command_prefix))
+            content='{}\n*To answer write* `{}answer "your message here"`'.
```
idodin: Is this a style change for yapf? It should not be breaking the build since it was passing before right? 

If its required then keep it, otherwise keep style the way it was before to keep a clean blame.

config/config.ini
```diff
@@ -49,6 +49,7 @@ GCWeatherURL = http://weather.gc.ca/city/pages/qc-147_metric_e.html
 GCWeatherAlertURL = https://weather.gc.ca/warnings/report_e.html?qc67
 WttrINTemplate = http://wttr.in/Montreal_2mpq_lang=en.png?_={}
 TepidURL = https://tepid.science.mcgill.ca:8443/tepid/screensaver/queues/status
+RecallChannel = food
```
idodin: Can we rename this to "FoodRecallChannel" to be more descriptive?

cogs/helpers.py
```diff
@@ -103,7 +152,7 @@ def __init__(self, bot):
             windchill_label = soup.find("a", string="Wind Chill")
             windchill = windchill_label.find_next().get_text().strip(
             ) + u"\xb0C"
-        except:
+        except Exception:
```
idodin: Nice catch on Base Exceptions being caught :)

cogs/helpers.py
```diff
@@ -35,15 +33,20 @@
...
-import os
 import datetime
+import pickle
+import feedparser
 from .utils.paginator import Pages
 from .utils.requests import fetch
 
 MCGILL_EXAM_URL = "https://www.mcgill.ca/exams/dates"
+
+CFIA_FEED_URL = "http://inspection.gc.ca/eng/1388422350443/1388422374046.xml"
```
idodin: Should this URL be configurable? The hash code or whatever these numbers are make me kind of nervous that the feed changes regularly....

cogs/helpers.py
```diff
@@ -53,11 +56,57 @@
 CURRENCY_TEMPLATE = "http://www.xe.com/currencyconverter/convert/" \
                     "?Amount={}&From={}&To={}"
 
+try:
```
idodin: This should be factored into the initialization of a separate cog for this functionality. (Another reason for not mixing this with the rest of helper functionality).

cogs/helpers.py
```diff
@@ -53,11 +56,57 @@
...
+    os.mkdir('./pickles')
+except Exception:
+    pass
+
 
 class Helpers():
     def __init__(self, bot):
         self.bot = bot
 
+    async def cfia_rss(self):
```
idodin: Can you explain this functionality to me? We're caching the last read id into a pickle, this co-routine periodically polls the XML for the latest recall id, and if that is not the cached ID, we send out a recall warning - correct? 

What happens in the case that we get 2 or more new recalls into the feed before refreshing? Isn't it better to increase the refresh timer and parse all missed recalls? - We reduce the risk of missing recalls, and its not critical for us to have recall warnings being posted with minute sensitivity, no? Overall, I think it also reduces load. 

If spamming multiple messages at a time is a concern, we can add a sleep timer to space them out too.

cogs/helpers.py
```diff
@@ -53,11 +56,57 @@
 CURRENCY_TEMPLATE = "http://www.xe.com/currencyconverter/convert/" \
                     "?Amount={}&From={}&To={}"
 
+try:
+    os.mkdir('./pickles')
```
idodin: What is the current state of logging? It might be time for us to make periodic backups of assets like this and logs. (Not a change for you to make, but something for us to think about).

cogs/reminder.py
```diff
@@ -181,7 +181,7 @@ def __init__(self, bot):
                         await asyncio.sleep(1)
 
             conn.close()
-            await asyncio.sleep(60)
+            await asyncio.sleep(60)    # seconds
```
idodin: In general, changes like this should be kept to a separate commit (as with the style change above) - the ideal case is for us to be able to find merge commits that identify exactly what has been introduced into master and check them out one by one to categorically see changes for each functionality. 

You don't have to change it, just a note :)

cogs/helpers.py
jidicula: It doesn't change, they just have multiple feeds for allergen contamination and location, but I picked the nationwide feed for food recalls.

cogs/mod.py
jidicula: I had been YAPFing with a different installation from the one installed in the repo by mistake, so this is fixing that incorrect YAPFing.

config/config.ini
jidicula: Sure thing

cogs/mod.py
idodin: Alright, great!

cogs/helpers.py
```diff
@@ -503,3 +552,4 @@ def __init__(self, bot):
 
 def setup(bot):
     bot.add_cog(Helpers(bot))
+    bot.loop.create_task(Helpers(bot).cfia_rss())
```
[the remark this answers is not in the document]
jidicula: Let's table this for now just so we can get the functionality into production, but I think moving forward we can have a new class for polling-based functions including food recalls, weather warnings, and métro service interruptions.

cogs/helpers.py
jidicula: I'll change the functionality to generate a dictionary of all current recall notices that aren't in the pickle if it exists. Then, each new notice would be posted to the server and the new recalls would be added to the dictionary, which would be pickled. This whole routine could be run daily.

cogs/helpers.py
jidicula: This will be addressed in a future PR once we have more subscription-based functions that aren't called by commands.

Issue #140, opened 2019-08-25
[BUG] [Easy fix] Bug with the new quoter delete function


le-potate:
During the 2 minute window that a quoter can delete, if he reacts with the delete emoji on any message it will delete the quote.

le-potate:
done

PR #141, opened 2019-08-25
Issue 140, fixes to quoter delete function

le-potate:
See #140 
The delete react now only works on the "Quote added." message , and it prevents Canary from deleting his own new quotes. Tested with my bot.


le-potate:
Added comments

PR #152, opened 2019-09-25
Adding food command for foodspotting

### LINKED ISSUE #151: Make marty send food spotting
Users can call `?food` to add food to the foodspotting channel
(Discussed in #idoneam)

le-potate:
Resolves #151 
This adds a food command that posts messages to the food spotting channel. Also, channels are now all taken by name in config.ini

### DIFF
--- a/cogs/helpers.py
+++ b/cogs/helpers.py
@@ -510,12 +510,16 @@ class Helpers(commands.Cog):
         except ValueError:
             ctx.send("Input must be integers")
 
-    @commands.command()
-    async def food(self, ctx, *args):
+    @commands.command(
+        aliases=['foodspot', 'fs', 'food', 'foodspotting', 'food_spotting'])
+    async def food_spot(self, ctx, *args):
         """Posts a food sale in #foodspotting.
-        Use: `?food Samosas in leacock`
-        You can also attach one picture to your message"""
-        message = " ".join(args)
+        Use: `?foodspot Samosas in leacock`
+        You can also attach one picture to your message (Write the command in the uploaded image caption)"""
+        if not args:
+            message = "\u200b"
+        else:
+            message = "**{}**".format(" ".join(args))
         channel = utils.get(self.bot.get_guild(
             self.bot.config.server_id).text_channels,
                             name=self.bot.config.food_spotting_channel)
@@ -524,15 +528,14 @@ class Helpers(commands.Cog):
         embed = discord.Embed()
         try:
             embed.set_image(url=ctx.message.attachments[0].url)
-        except:
+        except Exception:
             pass
         embed.set_footer(
             text=
-            "Added by {0} • Use '{1}food' if you spot food (See '{1}help food')"
+            "Added by {0} • Use '{1}foodspot' or {1}fs if you spot food (See '{1}help foodspot')"
             .format(username, self.bot.config.command_prefix[0]),
             icon_url=pfp)
-        embed.add_field(name="`Food spotted`", value="**{}**".format(message))
-
+        embed.add_field(name="`Food spotted`", value=message)
         await channel.send(embed=embed)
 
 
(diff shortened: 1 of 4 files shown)


cogs/helpers.py
```diff
@@ -510,6 +510,31 @@ def __init__(self, bot):
         except ValueError:
             ctx.send("Input must be integers")
 
+    @commands.command()
+    async def food(self, ctx, *args):
```
jidicula: Can you change the command name to something more related to its function, for example ?foodspot

cogs/helpers.py
```diff
@@ -510,6 +510,31 @@ def __init__(self, bot):
...
+        """Posts a food sale in #foodspotting.
+        Use: `?food Samosas in leacock`
+        You can also attach one picture to your message"""
```
jidicula: Does the command have to be added in the uploaded image caption or can the image be added in the next message after the command?

cogs/helpers.py
```diff
@@ -510,6 +510,31 @@ def __init__(self, bot):
...
+        channel = utils.get(self.bot.get_guild(
+            self.bot.config.server_id).text_channels,
+                            name=self.bot.config.food_spotting_channel)
+        username = ctx.message.author
+        pfp = ctx.message.author.avatar_url
+        embed = discord.Embed()
+        try:
+            embed.set_image(url=ctx.message.attachments[0].url)
+        except:
+            pass
```
jidicula: thank you for not leaving a bare `except`

cogs/mod.py
```diff
@@ -49,9 +51,10 @@ def __init__(self, bot):
         dest = user
         await dest.send(
             content='{}\n*To answer write* `{}answer "your message here"`'.
-            format(message, self.bot.config.command_prefix))
-        channel_to_forward = self.bot.get_channel(
-            self.bot.config.reception_channel_id)
+            format(message, self.bot.config.command_prefix[0]))
```
davidlougheed: ok as an aside why is this an array in the first place? can there be more than one?

cogs/mod.py
```diff
@@ -34,8 +35,9 @@ def __init__(self, bot):
             message = "{}".format(
                 " ".join(args)
             )    # to work regardless of whether the person uses apostrophes
-            channel_to_send = self.bot.get_channel(
-                self.bot.config.reception_channel_id)
+            channel_to_send = utils.get(self.bot.get_guild(
```
jidicula: this should be in another issue branch and another commit. please follow [atomic commit](https://en.wikipedia.org/wiki/Atomic_commit) guidelines.

cogs/mod.py
le-potate: it seems so, but also when you send multiple pictures it always sends them in different messages so i'm not sure when this happens

config/config.ini
```diff
@@ -28,7 +28,7 @@ UpvoteEmoji = upmartlet
 DownvoteEmoji = downmartlet
 ModeratorRole = Discord Moderator
 DeveloperRole = idoneam
-ReceptionChannelID = 454061583874785280
+ReceptionChannel = martys_dm
```
jidicula: atomic commits smh

cogs/helpers.py
jidicula: This should be specified in the help docstring.

le-potate: The command has to be added in the uploaded image caption

cogs/helpers.py
jidicula: wait jk that *is* a bare except. you can instead put

```python
except Exception:
    pass:
```

config/config.ini
le-potate: will be careful next time

cogs/helpers.py
jidicula: actually a shorter command name is better for mobile users

jidicula: e.g. ?fs

cogs/mod.py
jidicula: the iOS app allows multiple pics per message. not sure about android. definitely not an option on desktop.

Issue #153, opened 2019-09-25
Use embeds for quotes


jidicula:
### Description of Problem / Inefficiency:
Quotes is kind of janky for escaping pings.

### Implementation Suggestion:

Use Discord's embed feature to format quotes nicely. First line can be the date of the quote, 2nd line can be the content, and the bottom of the embed pane can have the quotee along with their avatar.

### Additional Context / Screenshots:
https://leovoel.github.io/embed-visualizer/

dbenjaminmiller:
Seems done. For anything more, open a new issue.

PR #155, opened 2019-09-27
Issue 153

jidicula:
No more pings in message contents, no need to escape usernames. Life
is good.

## Description
Uses Discord embed API to pretty-print quotes.

## Motivation and Context
Solves problem of pings in quote content and author fields.
Resolves: #153

## How Has This Been Tested?
Ran in main server with a database snapshot to most closely imitate production conditions.

## Types of changes

## Checklist:

### DIFF
--- a/cogs/quotes.py
+++ b/cogs/quotes.py
@@ -232,10 +232,9 @@ class Quotes(commands.Cog):
             for i, quote in zip(range(len(quote_list)), quote_list)
         ]
 
-        p = Pages(
-            ctx,
-            item_list=quote_list_text,
-            title='Quotes from {}'.format(quote_author.display_name))
+        p = Pages(ctx,
+                  item_list=quote_list_text,
+                  title='Quotes from {}'.format(quote_author.display_name))
 
         await p.paginate()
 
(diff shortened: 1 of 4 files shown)


cogs/quotes.py
```diff
@@ -199,9 +199,13 @@ def check(reaction, user):
         # current nick will be displayed, otherwise use the name stored
         # in db
 
-        author_name_unesc = author.display_name if author else name
```
le-potate: I think doing author.mention could be better

le-potate: oop i selected the wrong line, but it's the same for the new one

le-potate: So `author_name = author.mention if author else name`

le-potate: We have discovered it is not possible to mention users in author_name, only the ID appears

config/config.default.ini
```diff
@@ -49,7 +49,7 @@ GCWeatherURL = http://weather.gc.ca/city/pages/qc-147_metric_e.html
 GCWeatherAlertURL = https://weather.gc.ca/warnings/report_e.html?qc67
 WttrINTemplate = http://wttr.in/Montreal_2mpq_lang=en.png?_={}
 TepidURL = https://tepid.science.mcgill.ca:8443/tepid/screensaver/queues/status
-FoodRecallChannel = food
+FoodRecallChannel = bottest
```
le-potate: 👀

Issue #164, opened 2019-10-03
[BUG] embed images in quotes


jidicula:
### Bug Description
Not showing image URLs in quotes properly

### Reproduction Steps
Fetch any quote containing an image URL

### Potential Reasons behind Bug:
We switched to embeds for quotes in #156.

davidlougheed:
regex is too broad, re-opening

PR #165, opened 2019-10-11
Embed image URL in quote if it exists

jidicula:
## Description
Fixing a regression from PR #156. Image regex is conservative for jpg,
gif, png, and jpeg filename suffixes.

## Motivation and Context
Resolves: Issue #164 

## How Has This Been Tested?
Tested in test server for correct embedding and against false positives.

## Types of changes

## Checklist:


jidicula:
Yeah, and the embed panel doesn't allow multiple images per embed object.

dbenjaminmiller:
Exactly what I mean. The only thing would be to stitch them together but that's really overdoing it (and then you have to deal with different widths etc)

PR #184, opened 2019-10-18
Issue 136 Add new cog, roles and inrole commands

lamminade:
## Description
Added a new cog titled Info.py, this can be renamed later - mostly just wanted somewhere to put the new server info related functionality for roles, member info, etc.

Also added two role related functions:
- `inrole` : which lists all users who currently have the specified role in *paginated* form
- `roles` : lists all roles in server in *paginated* form, in the order listed in settings (highest permissions to lowest)

## Motivation and Context
This is motivated by issue 136, as well as the project reducing reliance on mosebot. 

## How Has This Been Tested?
Has been tested using large inputs, no inputs, etc. 

## Types of changes

- [?] Breaking change (fix or feature that would cause existing functionality to not work as expected)
(you never know)

## Checklist:

- [?] My change requires a change to the documentation.

### DIFF
--- a/cogs/info.py
+++ b/cogs/info.py
@@ -34,6 +34,7 @@ class Info(commands.Cog):
     async def roles(self, ctx):
         """Returns list of all roles in server"""
         roleNames = list(map(lambda role: role.name + "\n", ctx.guild.roles))
+        roleNames.reverse()
         p = Pages(ctx,
                   item_list=roleNames,
                   title="All roles in server",
--- a/cogs/info.py
+++ b/cogs/info.py
@@ -25,9 +25,22 @@ import asyncio
 from .utils.paginator import Pages
 import math
 
+
 class Info(commands.Cog):
     def __init__(self, bot):
         self.bot = bot
+
+    @commands.command()
+    async def roles(self, ctx):
+        """Returns list of all roles in server"""
+        roleNames = list(map(lambda role: role.name + "\n", ctx.guild.roles))
+        p = Pages(ctx,
+                  item_list=roleNames,
+                  title="All roles in server",
+                  display_option=(3, 20),
+                  editable_content=False)
+
+        await p.paginate()
 
     @commands.command()
     async def inrole(self, ctx, *, r):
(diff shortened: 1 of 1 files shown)


cogs/info.py
```diff
@@ -0,0 +1,72 @@
...
+    @commands.command()
+    async def inrole(self, ctx, *, r):
+        """Returns list of users in the specified role"""
+        members = []
+        for role in ctx.guild.roles:
+            if role.name.lower() == r.lower():
+                members = role.members
+                break
+
+        if (len(members) == 0): return
```
davidlougheed: There are some roles that exist but have 0 members, which are different from roles that don't exist – it might be good to display something for empty (but existing) roles

cogs/info.py
```diff
@@ -0,0 +1,72 @@
...
+
+        names = list(map(lambda m: str(m) + "\n", members))
+        numNames = len(names)
```
davidlougheed: This is only used once for the format below it, so maybe just use len(names) directly in the format string?

cogs/info.py
davidlougheed: Suggestion: add a flag `found` and return `if not found:`, or start with `members = None` and return `if members is None:`

cogs/info.py
```diff
@@ -0,0 +1,72 @@
...
+        """Returns list of all roles in server"""
+        roleNames = list(map(lambda role: role.name + "\n", ctx.guild.roles))
+        roleNames.reverse()
+        p = Pages(ctx,
+                  item_list=roleNames,
+                  title="All roles in server",
+                  display_option=(3, 20),
+                  editable_content=False)
+
+        await p.paginate()
```
jidicula: In general I'm averse to one-letter names except for indexing – could you give it a more descriptive name?

cogs/info.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @commands.command()
+    async def inrole(self, ctx, *, r):
```
jidicula: Same thing here as above with `r`

cogs/info.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class Info(commands.Cog):
+    def __init__(self, bot):
+        self.bot = bot
+
+    @commands.command()
+    async def roles(self, ctx):
+        """Returns list of all roles in server"""
+        roleNames = list(map(lambda role: role.name + "\n", ctx.guild.roles))
```
jidicula: Great use of `map` and `lambda`!

cogs/info.py
lamminade: I was following the convention used in quotes.py [line 270](https://github.com/idoneam/Canary/blob/cb71eeb6132d5ee119005398937278aca008bbf9/cogs/quotes.py#L270)

jidicula: I think we'll need to do this in the task/cleanup branch.

PR #200, opened 2019-11-01
Add manual windchill calculation

le-potate:
Environment Canada doesn't always give a value for wind chill but we are able to compute it ourselves instead. The formula used by Environment Canada is available [here](https://en.wikipedia.org/wiki/Wind_chill#North_American_and_United_Kingdom_wind_chill_index)
## Types of changes

## Checklist:

### DIFF
--- a/cogs/helpers.py
+++ b/cogs/helpers.py
@@ -131,63 +131,80 @@ class Helpers(commands.Cog):
     async def weather(self, ctx):
         """Retrieves current weather conditions.
         Data taken from http://weather.gc.ca/city/pages/qc-147_metric_e.html"""
+        def retrieve_string(label):
+            return soup.find(
+                "dt", string=label).find_next_sibling().get_text().strip()
+
         await ctx.trigger_typing()
 
         r = await fetch(self.bot.config.gc_weather_url, "content")
 
         soup = BeautifulSoup(r, "html.parser")
         # Get date
-        observed_label = soup.find("dt", string="Date: ")
+        observed_string = retrieve_string("Date: ")
         # Get temperature
-        temperature_label = soup.find("dt", string="Temperature:")
+        temperature_string = retrieve_string("Temperature:")
         # Get condition
-        condition_label = soup.find("dt", string="Condition:")
+        condition_string = retrieve_string("Condition:")
         # Get pressure
-        pressure_label = soup.find("dt", string="Pressure:")
+        pressure_string = retrieve_string("Pressure:")
         # Get tendency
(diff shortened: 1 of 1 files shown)


cogs/helpers.py
```diff
@@ -148,14 +148,19 @@ def __init__(self, bot):
...
-            windchill = u"N/A"
+        # Get windchill using standard formula from Environment Canada
+        temperature = float(
+            re.search(r"-?\d+\.\d",
+                      temperature_label.find_next_sibling().get_text().strip()
+                      ).group())
+        wind_speed = float(
+            re.search(
+                r"\d+",
+                wind_label.find_next_sibling().get_text().strip()).group())
```
jidicula: Since you're parsing the HTML here to get temperature and wind speed, you can pass these variables to the embed below instead of those values being filled by parsing again. You might have to cast them as `str` though, so their new values would be something like `value=str(temperature)` and `value=str(wind_speed)` when constructing the `Embed` object.

le-potate: They aren't parsed the same way though. For example, the string that is sent to embed is "WSW 68 gust 93 km/h", while in my wind_speed variable it is regexed to get only 68

jidicula: In that case then I suggest creating a new variable just to store the wind velocity string, and then you can regex that string just for the wind speed and the wind velocity string can be passed to the Embed object. This will reduce the number of parsing operations to get the same string. The wind velocity string is needed for both values: 1) it is regex-filtered for the windchill calculation, and 2) it is passed directly to the Embed.

The same thing could be done for temperature too.

cogs/helpers.py
```diff
@@ -148,14 +148,19 @@ def __init__(self, bot):
...
         wind_label = soup.find("dt", string="Wind:")
-        # Get windchill, only if it can be found.
-
-        try:
-            windchill_label = soup.find("a", string="Wind Chill")
-            windchill = windchill_label.find_next().get_text().strip(
-            ) + u"\xb0C"
-        except Exception:
-            windchill = u"N/A"
+        # Get windchill using standard formula from Environment Canada
```
davidlougheed: add link to formula so we can see the constants

davidlougheed:
image

le-potate:
As commented in the code, I now use the formula from MetService, which will give a Feels like temperature for any temperature. (For temperatures of 10 degree celcius and less, it actually uses the Environment Canada formula). See https://blog.metservice.com/FeelsLikeTemp

davidlougheed:
needs 80 character lines - i can do this if you want

Issue #221, opened 2019-11-29
[BUG] Metro status throws an error for line colour sometimes


davidlougheed:
Stack trace:

```
Nov 28 22:08:26 amnesiac python3[31944]: Traceback (most recent call last):
Nov 28 22:08:26 amnesiac python3[31944]:   File "/home/davidlougheed/marty/Canary/cogs/subscribers.py", line 133, in metro_status
Nov 28 22:08:26 amnesiac python3[31944]:     colour=line_colour)
Nov 28 22:08:26 amnesiac python3[31944]:   File "/home/davidlougheed/.local/share/virtualenvs/Canary-9hgJa66m/lib/python3.6/site-packages/discord/embeds.py", line 113, in __init__
Nov 28 22:08:26 amnesiac python3[31944]:     self.colour = colour
Nov 28 22:08:26 amnesiac python3[31944]:   File "/home/davidlougheed/.local/share/virtualenvs/Canary-9hgJa66m/lib/python3.6/site-packages/discord/embeds.py", line 210, in colour
Nov 28 22:08:26 amnesiac python3[31944]:     raise TypeError('Expected discord.Colour, int, or Embed.Empty but received %s instead.' % value.__class__.__name__)
Nov 28 22:08:26 amnesiac python3[31944]: TypeError: Expected discord.Colour, int, or Embed.Empty but received str instead.
```

PR #222, opened 2019-12-01
[fix][subscribers] Regenerate metro status dict

jidicula:
The metro status dictionary was not being regenerated correctly,
leading to type errors. There was also an incorrect comparison being
made in the logical conditions that would trigger a status update
being sent and saved.

**What this change does**
Refactors globals for default status messages and metro
colours. Refactors metro status dict and fixes conditional comparison
for triggering a status being posted in the server and updated in the dict.

**Any side-effects?**
None.

**Additional context/notes/links**
Note that this could not be adequately tested as metro outages are
infrequent.

Thanks to @davidlougheed for the refactor work.

Resolves: #221 

## Types of changes

## Checklist:

### DIFF
--- a/cogs/subscribers.py
+++ b/cogs/subscribers.py
@@ -36,14 +36,27 @@ import requests
 CFIA_FEED_URL = "http://inspection.gc.ca/eng/1388422350443/1388422374046.xml"
 METRO_STATUS_API = "https://www.stm.info/en/ajax/etats-du-service"
 
+METRO_GREEN_LINE = "1"
+METRO_ORANGE_LINE = "2"
+METRO_YELLOW_LINE = "4"
+METRO_BLUE_LINE = "5"
+
+METRO_COLOURS = {
+    METRO_GREEN_LINE: 0x008E4F,
+    METRO_ORANGE_LINE: 0xF08123,
+    METRO_YELLOW_LINE: 0xFFE400,
+    METRO_BLUE_LINE: 0x0083CA,
+}
+
+METRO_NORMAL_SERVICE_MESSAGE = "Normal métro service"
+
 # Default values by line number for status
-# Integers in list are line colours
 metro_status = {
-    "1": ["Normal métro service", 36431],    # Green Line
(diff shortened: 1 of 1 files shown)


cogs/subscribers.py
```diff
@@ -122,11 +135,11 @@ def check_status(line_number, response):
             response = requests.request("GET", METRO_STATUS_API)
             for line_status in metro_status.items():
                 line_number = line_status[0]
-                cached_status = line_status[1][0]
-                line_colour = line_status[1][1]
+                cached_status = line_status[1]
+                line_colour = METRO_COLOURS[line_number]
```
davidlougheed: this variable is probably unnecessary since it's just used once

cogs/subscribers.py
```diff
@@ -122,11 +135,11 @@ def check_status(line_number, response):
             response = requests.request("GET", METRO_STATUS_API)
```
davidlougheed: use `requests.get` instead

cogs/subscribers.py
```diff
@@ -122,11 +135,11 @@ def check_status(line_number, response):
...
             for line_status in metro_status.items():
                 line_number = line_status[0]
-                cached_status = line_status[1][0]
-                line_colour = line_status[1][1]
+                cached_status = line_status[1]
+                line_colour = METRO_COLOURS[line_number]
                 current_status = check_status(line_number, response)
```
davidlougheed: expand out tuple so it makes more sense and lines up with the other `_status` variable, e.g.:
```python
line_name, current_status = check_status(line_number, response)
```

cogs/subscribers.py
jidicula: Resolved by d13c353.

cogs/subscribers.py
jidicula: Resolved by d13c353.

cogs/subscribers.py
jidicula: Resolved by d13c353.

PR #242, opened 2020-01-01
Switch Dependency Management to Poetry

### LINKED ISSUE #241: Switch Dependency Management to Poetry [Enhancement]
We're currently using pipenv for dependency management. Pipenv seems to be acting up fairly consistently, particularly when locking dependencies. While, ideally, we'd investigate this further Poetry seems to be more appropriate (as pipenv's main benefit over poetry is python version management, which we don't need with #211 (Dockerization) ).

Port Pipenv and Pipenv.lock into Poetry's pyproject.toml . 

### Additional Context / Screenshots:
[Benchmark Comparison](https://johnfraney.ca/posts/2019/03/06/pipenv-poetry-benchmarks-ergonomics/)

idodin:
Ports dependency management from Pipenv to Poetry and updates relevant
README sections.
Updates Travis settings accordingly.

## Description
Adds pyproject.toml and poetry.lock to port from Pipenv to Poetry, updating README with relevant changes.
Details about dependencies remain unchanged (i.e. one-to-one port from what was in pipfile).
Updates Travis to use Poetry instead of Pipenv.

## Motivation and Context
Poetry is more consistent than Pipenv, especially when locking which is frustrating for a lot of developers. 
Resolves #241

## How Has This Been Tested?
Manual test on System Python and PyEnv Python. It seems like Poetry will pick up the specified Python version if it is found in the system (either system default or in pyenv list) (plays nicely with PyEnv). 

## Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


krubenok:
Looks like [Dependabot also supports Poetry](https://dependabot.com/blog/announcing-poetry-support/) so we're good on that front!

pyproject.toml
```diff
@@ -0,0 +1,26 @@
...
+readme = "README.md"
+authors = []
+
+[tool.poetry.dependencies]
+python = "^3.6.0"
+beautifulsoup4 = "==4.8.0"
+sympy = "==1.5"
+requests = ">=2.20.0"
+tabulate = "==0.8.6"
+"discord.py" = "==1.2.5"
```
jidicula: Is there a reason why `discord.py` is enclosed in quotes?

idodin: As mentioned on discord, poetry doesn't like dots. Enclosing the dependency name forces `discord.py` to be read as a dependency as opposed to a property.

jidicula: Cool, thanks for explaining!

jidicula:
@idodin is this good to merge?

idodin:
Sorry for the MIA - Pipfiles caused conflict. Removed and merged :)

PR #288, opened 2020-04-01
Self Assign Roles

### LINKED ISSUE #136: Role-related functionality
### Feature Description:
Make a substitute for .inrole, etc. to reduce nadeko dependency. I can probably do it myself when I get around to it.

I'll read the `discord.py` docs later.

krubenok:
## Description

I have created basic roles functionality to allow users to self-assign roles to themselves in certain categories. I have also relocated other role related functionality in the project to `roles.py` as it seemed better located there with other similar functions. In addition to the `?iam` generic role command, some commands have been created with limitations such as year and faculty (where a user may have only one role in that category at a time). 

## Motivation and Context

A lot of mod labour is wasted on assigning roles of little importance that are mostly just for community convenience. Letting them be self assigned within reason should save a decent amount of work. 

Related to #136 

## How Has This Been Tested?

Tested in the testing server. Would appreciate more tests with more roles to check for more corner cases. 

## Types of changes

## Checklist:

### DIFF
--- a/cogs/roles.py
+++ b/cogs/roles.py
@@ -38,6 +38,9 @@ class Roles(commands.Cog):
         """
         member = ctx.message.author
 
+        # normalize pronoun
+        pronoun = pronoun.title()
+
         pronouns = self.roles.get("pronouns").split(", ")
 
         if (transaction == "add") and (pronoun in pronouns):
--- a/cogs/roles.py
+++ b/cogs/roles.py
@@ -61,6 +61,9 @@ class Roles(commands.Cog):
         """
         member = ctx.message.author
 
+        # normalize field
+        field = field.title()
+
         fields = self.roles.get("fields").split(", ")
 
         if (transaction == "add") and (field in fields):
@@ -79,6 +82,9 @@ class Roles(commands.Cog):
         Self-assign a faculty of study role to a user. 
         """
         member = ctx.message.author
+
+        # normalize field
+        faculty = faculty.title()
 
         faculties = self.roles.get("faculties").split(", ")
 
@@ -107,6 +113,9 @@ class Roles(commands.Cog):
         Self-assign a year of study role to a user. 
         """
         member = ctx.message.author
+
+        # normalize year
+        year = year.title()
 
         years = self.roles.get("years").split(", ")
 
--- a/cogs/roles.py
+++ b/cogs/roles.py
@@ -1,0 +1,144 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) idoneam (2016-2019)
+#
+# This file is part of Canary
+#
+# Canary is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Canary is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Canary. If not, see <https://www.gnu.org/licenses/>.
+
+# discord-py requirements
+import discord
+from discord import utils
+from discord.ext import commands
+
+import codecs
+import configparser
+
+
+class Roles(commands.Cog):
+    def __init__(self, bot):
+        self.bot = bot
+        self.roles = self.bot.config.roles
+
+    @commands.command()
+    async def pronoun(self, ctx, transaction, pronoun):
+        """
+        Self-assign a pronoun role to a user. 
+        """
+        member = ctx.message.author
+
+        pronouns = self.roles.get("pronouns").split(", ")
+
+        if (transaction == "add") and (pronoun in pronouns):
+            role = discord.utils.get(ctx.guild.roles, name=pronoun)
+
+            # Allow only a single pronoun role per user.
+            for item in pronouns:
+                role_item = discord.utils.get(ctx.guild.roles, name=item)
+                if role_item is not None:
+                    await member.remove_roles(role_item,
+                                              reason="Self Requested")
+
+            await member.add_roles(role, reason="Self Requested")
+        elif (transaction == "remove") and (pronoun in pronouns):
+            role = discord.utils.get(ctx.guild.roles, name=pronoun)
+            await member.remove_roles(role, reason="Self Requested")
+
+    @commands.command()
+    async def field(self, ctx, transaction, field):
+        """
+        Self-assign a field of study role to a user. 
+        """
+        member = ctx.message.author
+
+        fields = self.roles.get("fields").split(", ")
+
+        if (transaction == "add") and (field in fields):
+            role = discord.utils.get(ctx.guild.roles, name=field)
+            await member.add_roles(role, reason="Self Requested")
+
+        elif (transaction == "remove") and (field in fields):
+            role = discord.utils.get(ctx.guild.roles, name=field)
+            await member.remove_roles(role, reason="Self Requested")
+
+    @commands.command()
+    async def faculty(self, ctx, transaction, faculty):
+        """
+        Self-assign a faculty of study role to a user. 
+        """
+        member = ctx.message.author
+
+        faculties = self.roles.get("faculties").split(", ")
+
+        if (transaction == "add") and (faculty in faculties):
+            role = discord.utils.get(ctx.guild.roles, name=faculty)
+
+            # Allow only a single faculty role per user.
+            for item in faculties:
+                role_item = discord.utils.get(ctx.guild.roles, name=item)
+                if role_item is not None:
+                    await member.remove_roles(role_item,
+                                              reason="Self Requested")
+
+            await member.add_roles(role, reason="Self Requested")
+
+        elif (transaction == "remove") and (faculty in faculties):
+            role = discord.utils.get(ctx.guild.roles, name=faculty)
+            await member.remove_roles(role, reason="Self Requested")
+
+    @commands.command()
+    async def year(self, ctx, transaction, year):
+        """
+        Self-assign a year of study role to a user. 
+        """
+        member = ctx.message.author
+
+        years = self.roles.get("years").split(", ")
+
+        if (transaction == "add") and (year in years):
+            role = discord.utils.get(ctx.guild.roles, name=year)
+
+            # Allow only a single year role per user.
+            for item in years:
+                role_item = discord.utils.get(ctx.guild.roles, name=item)
+                if role_item is not None:
+                    await member.remove_roles(role_item,
+                                              reason="Self Requested")
+
+            await member.add_roles(role, reason="Self Requested")
+
+        elif (transaction == "remove") and (year in years):
+            role = discord.utils.get(ctx.guild.roles, name=year)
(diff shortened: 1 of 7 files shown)


krubenok:
> 3\. There is no reason that the list of assignable roles couldn't be in the database, but I do suppose that that would take more effort and would be of relatively minimal benefit (essentially, it would be beneficial only in letting mods change roles and letting them do so without having to restart Canary or bother the bot's sysop).

This is the intention, but I've punted it to the role creation functionality for the time being it's gonna stick to the config file.

config/config.ini
```diff
@@ -79,3 +79,10 @@ TransactionTax = 0.15
 [Betting]
 RollCases = 66, 90, 99, 100
 RollReturns = 0, 2, 4, 10
+
+[Roles]
+Pronouns = She/Her, He/Him, They/Them
+Fields = Anatomy and Cell Biology, Anthropology, Agriculture, Architecture, Biochemistry, Bioengineering, Biology, Bioresource Engineering, Chemical Engineering, Chemistry, Civil Engineering, Classics, cogito, Commerce, Computer Engineering, Computer Science, Computer Science/Biology, Cultural Studies, Desautels, Economics , Electrical Engineering, English, Experimental Medicine, Finance, Geography, History, Indigenous Studies, International Development Studies, Jewish Studies, mac kid, Materials Engineering, Math, MBA, Mechanical Engineering, Medicine, Microbiology and Immunology, Neuroscience, Nursing, Pharmacology, Philosophy, Physical Therapy, Physics, Physiology, Political Science, Psychiatry, Psychology, Public Health, Social Work, Sociology, Software Engineering, Statistics
+Faculties = Science, Engineering, Management, Arts, ArtSci, Agriculture and Environment, Law, Education, Dentistry, Music
+Years = U0, U1, U2, U3, U4, PhD, Masters, Post Doc, wenthere
+Generics = weeb, weeb stomper
```
krubenok, line 88: This is just the default suggested set but the set used on our server can obviously differ from this.

cogs/roles.py
```diff
@@ -32,132 +32,89 @@ def __init__(self, bot):
         self.bot = bot
         self.roles = self.bot.config.roles
 
-    @commands.command()
-    async def pronoun(self, ctx, transaction, pronoun):
+    async def toggleRole(self, ctx, transaction, requestedRole, category):
         """
-        Self-assign a pronoun role to a user. 
+        Assigs a single role to a user with no checks from a category of roles
```
ValerianClerc, line 37: Typo: Assigns*

cogs/roles.py
```diff
@@ -32,132 +32,89 @@ def __init__(self, bot):
...
+        if (transaction == "add") and (requestedRole in roles):
+            role = discord.utils.get(ctx.guild.roles, name=requestedRole)
             await member.add_roles(role, reason="Self Requested")
-        elif (transaction == "remove") and (pronoun in pronouns):
-            role = discord.utils.get(ctx.guild.roles, name=pronoun)
+        elif (transaction == "remove") and (requestedRole in roles):
+            role = discord.utils.get(ctx.guild.roles, name=requestedRole)
             await member.remove_roles(role, reason="Self Requested")
-
         else:
```
ValerianClerc, line 51: Should we have an `elif` for a different error message sent back if the transaction is invalid (if it's not "add" or "remove")?

cogs/roles.py
```diff
@@ -32,132 +32,89 @@ def __init__(self, bot):
...
-            await ctx.send("Error, that role doesn't exist")
+            await ctx.send("Error, that role doesn't exist in this category")
+        return
 
-    @commands.command()
-    async def field(self, ctx, transaction, field):
+    async def toggleRestrictedRole(self, ctx, transaction, requestedRole,
```
ValerianClerc, line 55: I kind of like `toggleUniqueRole` a bit better because it seems clearer to me what the purpose of the function is. Up to you though

cogs/roles.py
krubenok: Do I need a different error for add/remove being wrong vs. the role not existing or is one catch all error sufficient?

cogs/roles.py
```diff
@@ -0,0 +1,163 @@
...
+                if old_role is not None:
+                    await member.remove_roles(old_role,
+                                              reason="Self Requested")
+            await member.add_roles(role, reason="Self Requested")
+        elif (transaction == "remove") and (requestedRole in roles):
+            role = discord.utils.get(ctx.guild.roles, name=requestedRole)
+            await member.remove_roles(role, reason="Self Requested")
+        else:
+            await ctx.send("Error, that role doesn't exist in this category")
+        return
```
davidlougheed, line 81: dont need this return

cogs/roles.py
ValerianClerc: I was thinking that it could be nice for the user of the command to know if they typed in a non-existant role or if they just used the command wrong. But yeah this is super subjective

cogs/roles.py
```diff
@@ -0,0 +1,163 @@
...
+                old_role = discord.utils.get(ctx.guild.roles,
+                                             name=category_roles)
+                if old_role is not None:
+                    await member.remove_roles(old_role,
+                                              reason="Self Requested")
+            await member.add_roles(role, reason="Self Requested")
+        elif (transaction == "remove") and (requestedRole in roles):
+            role = discord.utils.get(ctx.guild.roles, name=requestedRole)
+            await member.remove_roles(role, reason="Self Requested")
+        else:
```
davidlougheed, line 79: maybe this should early-return instead of being a final else
`if requestedRole not in roles`
different one for `transaction not in ("add", "remove")` or whatever for a different err msg like what val's talking abt

then we can move 
```python
role = discord.utils.get(ctx.guild.roles, name=requestedRole)
```
out of the blocks so that line isnt duplicated twice

cogs/roles.py
```diff
@@ -0,0 +1,163 @@
...
+    def __init__(self, bot):
+        self.bot = bot
+        self.roles = self.bot.config.roles
+
+    async def toggleRole(self, ctx, transaction, requestedRole, category):
+        """
+        Assigs a single role to a user with no checks from a category of roles
+        """
+        member = ctx.message.author
+        roles = self.roles.get(category).split(", ")
```
davidlougheed, line 40: here u can just index with `[]`
or, if we want to prevent key errors, we should do a default that is `""` instead of `None`
since if the get gives None as a default itll just raise exception on split anyway

config/config.ini
```diff
@@ -79,3 +79,10 @@ TransactionTax = 0.15
 [Betting]
 RollCases = 66, 90, 99, 100
 RollReturns = 0, 2, 4, 10
+
+[Roles]
+Pronouns = She/Her, He/Him, They/Them
```
davidlougheed, line 84: gonna hav eto clear all of these categories with mods sry

cogs/roles.py
```diff
@@ -0,0 +1,163 @@
...
+            await ctx.send("Error, that role doesn't exist in this category")
+        return
+
+    async def toggleRestrictedRole(self, ctx, transaction, requestedRole,
+                                   category):
+        """
+        Assigs a single role to a user with no checks from a category of roles
+        """
+        member = ctx.message.author
+        roles = self.roles.get(category).split(", ")
```
davidlougheed, line 61: same thing with the get()

cogs/roles.py
```diff
@@ -0,0 +1,163 @@
...
+        Self-assign a pronoun role to a user. 
+        """
+        await Roles.toggleRole(self, ctx, transaction, pronoun, "pronouns")
+
+    @commands.command()
+    async def field(self, ctx, transaction, field):
+        """
+        Self-assign a field of study role to a user. 
+        """
+        await Roles.toggleRole(self, ctx, transaction, field, "fields")
```
davidlougheed, line 95: not a static fn, call with `self.toggleRole` and skip passing the self arg by hand
same goes for all of these calls

config/config.ini
krubenok: Yep, goal is to get some in here so that we can test as the default config. The actual config will differ a lot.

cogs/roles.py
krubenok: Going to go with a more generic error message catch all here.

cogs/roles.py
```diff
@@ -0,0 +1,166 @@
...
+        # normalize user input to title case
+        requested_role = requested_role.title()
+        role = discord.utils.get(ctx.guild.roles, name=requested_role)
+
+        if (requested_role not in roles):
+            await ctx.send("Invalid Role")
+            return
+
+        if (transaction == "add"):
+            # Allow only a single faculty role per user.
```
jidicula, line 74: This comment could be made a bit more general since this function isn't going to be called just for faculty role assignment.

cogs/roles.py
```diff
@@ -0,0 +1,166 @@
...
+            return
+
+        if (transaction == "add"):
+            await member.add_roles(role, reason="Self Requested")
+        elif (transaction == "remove"):
+            await member.remove_roles(role, reason="Self Requested")
+        else:
+            await ctx.send("Must `add` or `remove` a role")
+
+    async def toggle_unique_role(self, ctx, transaction, requested_role,
```
jidicula, line 57: I'm not sure that "unique" is the right word for naming here – perhaps some use of "precluding" in the name would be more apt? I feel like there might be some set theory or discrete math term that describes this property exactly, but I'm not knowledgeable enough to know what term would be best.

cogs/roles.py
```diff
@@ -0,0 +1,166 @@
...
+                old_role = discord.utils.get(ctx.guild.roles,
+                                             name=category_roles)
+                if old_role is not None:
+                    await member.remove_roles(old_role,
+                                              reason="Self Requested")
+            await member.add_roles(role, reason="Self Requested")
+        elif (transaction == "remove"):
+            await member.remove_roles(role, reason="Self Requested")
+        else:
+            await ctx.send("Must `add` or `remove` a role")
```
jidicula, line 85: I see a lot of duplicated code between these 2 functions – is it possible to factor out the common parts, and functionalize the "precluding" operation in `toggle_unique_role()`?

Another option would be to write a general function for role addition and add an enclosing category around the faculty and year categories to denote them as single-assignment, then include a check in the function to see if the category argument is in the "single-assignment" category.

cogs/roles.py
```diff
@@ -0,0 +1,166 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Canary. If not, see <https://www.gnu.org/licenses/>.
+
+# discord-py requirements
+import discord
+from discord import utils
+from discord.ext import commands
+
+import codecs
+import configparser
```
jidicula, line 26: Almost forgot – these 2 imports are unused in this file, so they can probably be removed.

cogs/roles.py
krubenok: Restricted? idk asking set theory friends lol.

cogs/roles.py
```diff
@@ -0,0 +1,146 @@
...
+        Assigns a single role to a user with no checks from a category of roles
+        """
+        member = ctx.message.author
+        roles = self.roles[category].split(", ")
+
+        # normalize user input to title case
+        requested_role = requested_role.title()
+        role = utils.get(ctx.guild.roles, name=requested_role)
+
+        if (requested_role not in roles):
```
davidlougheed, line 48: ok now im getting pedantic but theres a lot of redundant parens in this diff
im fine with merging without changing these and ill do a cleanup later

Issue #236, opened 2019-12-21
[BUG] Need to handle thread crashes in subscribers


davidlougheed:
### Bug Description
Sometimes a subscriber with an unhandled exception will crash, and since it appears to be in another thread it won't get caught, resulting in the thread dying.

### Reproduction Steps
See issues with #stm_alerts

### Potential Reasons behind Bug:
See above.

### Screenshots:
![image](https://user-images.githubusercontent.com/357970/71310432-1deeed00-23e2-11ea-9a1c-225e39e64430.png)

PR #312, opened 2020-07-27
Refactor subscribers and handle errors better

davidlougheed:
## Description

* Improve performance of subscribers by caching some values in the cog instance
* Avoid creating cog multiple times
* Handle uncaught exceptions from the coroutines 
* Reduce repetition using a decorator

## Motivation and Context
* Issue #236 
* Making it easier to make new subscribers

## How Has This Been Tested?
* Made sure subscribers still ran

## Checklist:


cogs/utils/subscribers.py
```diff
@@ -0,0 +1,58 @@
+# Copyright (C) idoneam (2016-2020)
```
jidicula, line 1: should this file be named differently to avoid confusing its purpose with `cogs/subscribers.py`?

davidlougheed:
@le-potate will this conflict at all with the webhook logger stuff?

le-potate (approved):

Issue #47, opened 2018-11-14
Expand DB Structure


lazho:
# Enhancement Request: [Title]

### Description of Problem / Inefficiency:
DB is small, need to expand to keep track of more information about up/downmartlets.

### Implementation Suggestion:
See screenshots/message pastes

### Additional Context / Screenshots:
![image](https://user-images.githubusercontent.com/2417212/48497663-9dcf7e80-e802-11e8-8a01-b74162ade0df.png)

david:
```
i mean its still hitting the db what like 10 times in <1 second, which isnt huge

ok so there's also the compromise
store rows of (voter, votee, channel, up, down)  so we have the stats we want but reduce read/write a bit
wait arent we hitting the db a bunch anyway
since you have to update the score lol
i guess its smaller updates
```

ida:
```
we could do batch read/writes maybe
idk if that's actually possible at all with discord

nah but like if multiple people upvote something at the same time idk how well sqlite3 handles multiple open db connections/possibly overlapping writes
we could do a cache thing and every second or something move everything into the db
which is probably overkill lol
```

lazho:
```
but my concerns was simultaneous read/writes needing locks on the db
but i think for such a small app it should have absolutely no problem at all to store each vote action on one row
we should not denormalize/flatten the count
also like members should have primary key (obviously) and a unique constraint on a long which is the discord snowflake identifier for the user
should not be storing someone's display name in the db bc these change all the time
pretty sure caching the score would decrease performance from needing to update the score all the time
since the vote table is already foreign keying to members table it will have an index on the votee's id and can be quickly loaded/counted live
```

PR #319, opened 2020-09-01
Rewrite score completely: Log all reactions and get advanced stats for all emojis

### LINKED ISSUE #37: Score breakdown
Sometimes in addition to current scores, users may want to know how many upmartlets or downmartlets they have accumulated, plus some extra information.

- Keep upmartlet and downmartlet counts separately in the database, and showing the total score by calculating the difference.
- Show stats like "channel where you got the most up/downmartlets", etc. Thought it might be fun

le-potate:
`?score` and `?ranking` can now be used with every emoji. 
Usage:
-  `?score`: as before, returns the total score of the user who called it (upmartlets - downmartlets)
- `?score @user`: returns the total score of the user mentioned
- `?score :emoji:`: returns the number of times the user who called the command received that emoji as reaction
- `?score :emoji: @user` OR `?score @user :emoji:`: returns the number of times the user mentioned received that emoji as reaction
- `?ranking`: as before, returns the total score ranking of all users. (The users that appears are the ones who received at least one upmartlet or downmartlet. Before, this used to be all users that got added to the Members table)
- `?ranking :emoji:`: returns the ranking of the number of times users got this emoji as reaction

For all of these, self-reactions are ignored as before. Even though there is now a lot more calculations involved, the result is still nearly instant.

## Description
Adds a Reactions table that, on every raw reaction added, will keep the ID of the person that reacted (ReacterID),the ID of the person that wrote the message (ReacteeID), the name of the reaction (this is the result of str(emoji), that is, either the emoji directly for unicode directly, or <:emojiname:emojiid> for custom emoji), and the ID of the message. The ReacterID and ReacteeID reference IDs in the Members table. Each row is unique since a certain user can only react a certain emoji on a certain message once. On every raw reaction removed, the reaction is removed from the table. 

Since score is now calculated directly from that table, there is no need to cache score in the Members table anymore. DisplayName is renamed Name to avoid confusion since it will now use the username#descriminator instead (or if that is unavaible, the ID). The reference to DisplayName in currency.py is changed to Name.

On reaction add/remove, if the Reacter or Reactee users aren't in the Members table, they are added. If the user isn't in the server anymore, take a little bit more time to try and find their name in the rest of Discord (using fetch_user instead of get_user). If the account is deleted, use the ID as name.

If a user updates their username, it is updated in the database (this should possibly be moved out of score.py in the future)

**I have also built a database of all reactions that were used to date on the discord server (~350000), which we need to add to the Marty database at the next update. I have also rebuilt the `Members` database with all the current users and all users that ever got a reaction/reacted on a message. We must check that all users in `BankTransactions` are in this new `Members` table to not get problems since `UserID` in `BankTransactions` references `ID` in `Members`. However, I noticed while coding this that Foreign Keys don't seem work in the database, which needs to be fixed before merging this. When that is fixed, there will be problems with `currency.py` and the `BankTransactions` table because right now users are not inserted into `Members` when they are inserted into `BankTransactions`, which they should be for the Foreign Key to work.**

## Motivation and Context
Resolves #37
Part of #47, unsure if it resolves it?
Useful to get very interesting stats. 

## How Has This Been Tested?
All the commands were tested with all possible arguments and many invalid inputs. The raw reaction add/remove was tested with a couple of users and works. I tested self-reacting and it wasn't counted.

## In the future
I have written some demos which aren't completely ready yet, such as:
- How many reacts a user received (and the ranking for this for each user)
- How many reacts a user put on messages (and the ranking for this for each user)
- The ranking of reacts a user received most often
- The ranking of reacts a user put on messages most often
- The ranking of the most used emojis on the server
- How many reacts in total were received on the server

## Types of changes

## Checklist:


Martlet.schema
```diff
@@ -25,9 +25,8 @@ CREATE TABLE IF NOT EXISTS "Quotes" (
 	`Date`	TEXT
 );
 CREATE TABLE IF NOT EXISTS `Members` (
-	`ID`	      INTEGER UNIQUE,
-	`DisplayName` TEXT,
-	`Score`	      INTEGER
+	`ID`    INTEGER UNIQUE,
+	`Name`  TEXT
```
davidlougheed, line 29: ouch we'll need to do some weird stuff here

cogs/score.py
```diff
@@ -42,146 +43,217 @@ async def on_ready(self):
...
-                "upvote/downvote moar.")
-            return
 
-        table = []
-        table_list = []
-        counter = 1
+        if not emoji:
+            emoji = "total"
+            c.execute(
+                "SELECT count(ReacteeID) FROM Reactions WHERE ReacteeID = ? AND ReactionName = ? AND ReacterID != ReacteeID",
```
davidlougheed, line 165: this is over 80 chars, im surprised YAPF didnt yell

cogs/score.py
```diff
@@ -42,146 +43,217 @@ async def on_ready(self):
...
+                "FROM Reactions AS R, Members as M "
+                "WHERE R.ReactionName = ? "
+                "AND R.ReacterID != R.ReacteeID "
+                "AND R.ReacteeID = M.ID "
+                "GROUP BY R.ReacteeID "
+                "ORDER BY count(*) DESC", (str(self.DOWNMARTLET), ))
+            downmartlet_dict = dict(c.fetchall())
+            total_score_dict = {
+                key: upmartlet_dict.get(key, 0) - downmartlet_dict.get(key, 0)
+                for key in list(upmartlet_dict.keys()) +
```
davidlougheed, line 213: don't need the .keys() here
I would rewrite this as
```python
from itertools import chain
# ...
for key in chain(upmartlet_dict, downmartlet_dict)
```

this will iterate through all the keys without the overhead of putting all the keys into a list first

cogs/score.py
```diff
@@ -42,146 +43,217 @@ async def on_ready(self):
...
+                list(downmartlet_dict.keys())
+            }
+            sorted_score_dict = dict(Counter(total_score_dict).most_common())
+
+            names = list(
+                map(lambda x, y: f"{y}. {x}", sorted_score_dict.keys(),
```
davidlougheed, line 219: don't need .keys() again

cogs/score.py
```diff
@@ -42,146 +43,217 @@ async def on_ready(self):
...
-
-    async def _update_score_if_needed(self, payload, negated=False):
-        # Check for Martlet emoji + upmartletting yourself
+    async def _get_name_from_id(self, user_id):
+        try:
+            user = self.bot.get_user(user_id)
+            if user is None:
+                raise AttributeError
+            name = str(user)
+        except AttributeError:
```
davidlougheed, line 52: does this occur somewhere other than the if statement above? if so from what?

cogs/score.py
```diff
@@ -42,146 +43,217 @@ async def on_ready(self):
...
+            except discord.errors.NotFound:
+                name = str(user_id)
+        return name
+
+    async def _add_member_if_needed(self, user_id):
+        conn = sqlite3.connect(self.bot.config.db_path)
+        c = conn.cursor()
+        c.execute("SELECT Name FROM Members WHERE ID = ?", (user_id, ))
+        fetch = c.fetchall()
+        if len(fetch) == 0:
```
davidlougheed, line 67: if not c.fetchone():

cogs/score.py
```diff
@@ -42,146 +43,217 @@ async def on_ready(self):
...
+            c.execute(
+                "SELECT M.Name, count(*) "
+                "FROM Reactions AS R, Members as M "
+                "WHERE R.ReactionName = ? "
+                "AND R.ReacterID != R.ReacteeID "
+                "AND R.ReacteeID = M.ID "
+                "GROUP BY R.ReacteeID "
+                "ORDER BY count(*) DESC", (emoji, ))
+            counts = list(zip(*c.fetchall()))
+            if len(counts) == 0:
```
davidlougheed, line 239: if not counts:

davidlougheed (changes requested):

Martlet.schema
le-potate: Yeah. Like I mentionned in the PR description, we're gonna need to fix foreign keys which currently don't work, which will already break currency because users were added into it without being added to Members first, which need to be changed.

I have already built the new members table with all the current users and all users that ever got a reaction/reacted on a message, so when foreign keys are fixed, and once currency adds users to members first, we can also add the current users in the currency table to Members

cogs/score.py
le-potate: oops didn't notice, and yeah I think yapf doesn't yell for strings

cogs/score.py
le-potate: I don't think so, changed it to if statement in 8a67002

cogs/score.py
le-potate: Resolved in 0847cae

cogs/score.py
le-potate: Resolved in 0847cae

cogs/score.py
le-potate: Resolved in 0847cae

cogs/score.py
le-potate: Resolved in 0847cae

cogs/score.py
le-potate: Resolved in 0847cae

le-potate:
We're also gonna need to change the way ?backup is sent because the database will now be larger than 8mb

cogs/score.py
```diff
@@ -17,15 +17,146 @@
...
 from .utils.paginator import Pages
 
+f = open('data/premade/emoji.json', encoding="utf8")
+EMOJI = json.load(f)
+
+
+class TotalEmojiConverter(commands.Converter):
+    async def convert(self, ctx, argument):
+        try:
+            result = await PartialEmojiConverter().convert(ctx, argument)
```
davidlougheed, line 45:
```py
return await PartialEmojiConverter().convert(ctx, argument)
```

cogs/score.py
```diff
@@ -17,15 +17,146 @@
...
+        except commands.BadArgument:
+            if not any(argument in d.values() for d in EMOJI.values()):
+                raise commands.BadArgument(
+                    "Not in the current list of Discord Unicode Emojis "
+                    "and no Custom Emoji found")
+            result = argument
```
davidlougheed, line 51:
```py
return argument
```

cogs/score.py
```diff
@@ -17,15 +17,146 @@
...
+        return result
```
davidlougheed, line 52: remove, do the same stuff to the rest

cogs/score.py
```diff
@@ -42,147 +173,542 @@ async def on_ready(self):
...
+
+        # If there is nothing in the where_list, we must make sure that
+        # we still give a condition since this returns a string to place after
+        # starting a WHERE clause
+        if not where_list:
+            where_list = ["1 = 1"]
+
+        if prefix:
+            where_list = [f"{prefix}.{item}" for item in where_list]
+        where_str = ' AND '.join(where_list)
```
davidlougheed, line 323: double quotes for consistency

cogs/score.py
```diff
@@ -42,147 +173,542 @@ async def on_ready(self):
...
+                   f"FROM Reactions "
+                   f"WHERE {where_str} "
+                   f"GROUP BY ReactionName "), t)
+
+        counts = list(zip(*c.fetchall()))
+        if not counts:
+            await ctx.send(embed=discord.Embed(title="No results found"))
+            return
+        names = counts[0]
+        values = counts[1]
```
davidlougheed, line 603:
```py
names, values = counts
```
does this work or is counts longer than length 2?

davidlougheed (changes requested):
overall lgtm, its been tested a bunch so this is essentially just more syntaxy stuff ill reread it after my chiken is delivered

cogs/score.py
le-potate: Resolved in 95d666f

cogs/score.py
le-potate: Resolved in 95d666f

cogs/score.py
le-potate: Resolved in 95d666f

cogs/score.py
le-potate: Resolved in 95d666f

cogs/score.py
le-potate: Resolved in 95d666f

le-potate:
To speed up the next update, we have decided to merge this now, and to possibly review code style later

davidlougheed (approved):

PR #321, opened 2020-09-01
Issue #223 Add command to see the previous roles of a user that left and came back and to give them back

### LINKED ISSUE #223: A function to give back roles to returning users
### Feature Description:
A command that lets moderators give back all the roles returning users had before the last time they left

When users leave, save their roles in db. When they come back, the command `>giverolesback @user` sends an embed with the list of roles, and a moderator can click the ok react to give them back. (This could be implemented differently but the important part is to be able to see all the roles before adding them back)

I don't think the number of users leaving will take too much space in the db.

le-potate:
Resolves #223

Adds a `?previousroles` command, which returns the list of roles that a user had before leaving the server. There is an OK react on this list, which can be used by mods to give the roles back to that user. This is useful for users as they don't have to remember the roles they had before leaving, and this is very useful for mods as they don't have to manually add the roles back one by one

## Description
-Add PreviousRoles table. Everytime a user leaves, their roles are added to the PreviousRoles table. If they had previously left, the roles are overwritten by the roles they have now. (Unless they now have no roles, in which case the previous list is kept)
-Add previous_roles command (alias previousroles, giverolesback and rolesback) that returns the list of roles the user had before leaving.
-There is an OK react on the previous_roles list. If a Discord Moderator clicks the OK react, the roles are given back to the user.
-In the discord moderator logs, Marty adds the roles with a reason that says `(moderator's name)  used the previous_roles command".

## How Has This Been Tested?
Has been tested with:
-A user that never left (As expected, you get a message saying roles cannot be found)
-A user that left and came back but didn't have roles (As expected, you get a message saying roles cannot be found)
-A user that left and came back and had roles (As expected, you get the list of roles)
-A user that left with roles, came back, left again with no roles, then came back. (As expected, you get the old list of roles)
-A user that left with roles, came back, left again with different roles, then came back. (As expected, you only get the new roles)
For all of these, a non-moderator clicked the OK button and it didn't do anything, as expected, and then a moderator clicked the OK button, which successfully added the roles to the user.

## In the future
- This list could possibly be sent everytime a user comes back. There could also be a message like "Welcome back! Use the ?previousroles command to see your previous roles", etc.
- There as been some discussion about this but I don't think it's a good idea to give the roles back automatically, since there is many problems that could arise from that (For example, moderators could decide to remove a certain role from a certain user, then that user leaves while Marty is down, which causes the database to still have the roles from a previous time that this user left, back when they had the role)
- It could be nice for mods that there is also a Trash button next to the OK button that lets them remove some roles from the list, then add the remaining roles back. However, this would require some changes to `paginator.py` since it can only have one editable_emoji at a time right now.

## Types of changes

## Checklist:


cogs/helpers.py
```diff
@@ -563,6 +564,94 @@ async def colour(self, ctx, *, arg: str):
         fn = "{}.{}".format(match.group(1), ext)
         await ctx.send(file=discord.File(fp=buffer, filename=fn))
 
+    @commands.Cog.listener()
+    async def on_member_remove(self, member):
+        roles_id = []
+        for role in member.roles:
+            if role.name != "@everyone":
+                roles_id.append(role.id)
```
davidlougheed, line 572:
```python
roles_id = [role.id for role in member.roles if role.name != "@everyone"]
```

cogs/helpers.py
```diff
@@ -563,6 +564,94 @@ async def colour(self, ctx, *, arg: str):
...
+        if len(roles_id) > 0:
```
davidlougheed, line 573:
```python
if roles_id:
```

cogs/helpers.py
```diff
@@ -563,6 +564,94 @@ async def colour(self, ctx, *, arg: str):
...
+    @commands.command(aliases=['previousroles', 'giverolesback', 'rolesback'])
+    async def previous_roles(self, ctx, user: discord.Member):
+        """Show the list of roles that a user had before leaving, if possible.
+        A moderator can click the OK react on the message to give these roles back
+        """
+        conn = sqlite3.connect(self.bot.config.db_path)
+        c = conn.cursor()
+        roles_id_string = c.execute(
+            'SELECT Roles FROM PreviousRoles WHERE ID = ?',
+            (user.id, )).fetchall()
```
davidlougheed, line 590: use fetchone instead of fetchall, roles_id_string is a confusing name because its not a string, its a tuple of one string

cogs/helpers.py
```diff
@@ -563,6 +564,94 @@ async def colour(self, ctx, *, arg: str):
...
+        if len(roles_id_string) > 0:
```
davidlougheed, line 591: assuming fetchone:
```python
if roles_id_string is not None:
```

cogs/helpers.py
```diff
@@ -563,6 +564,94 @@ async def colour(self, ctx, *, arg: str):
...
+            roles_id = list(roles_id_string[0][0].split(" "))
```
davidlougheed, line 592:
```python
roles_id = roles_id_string[0].split(" ")
```
the list cast is redundant for split

cogs/helpers.py
```diff
@@ -563,6 +564,94 @@ async def colour(self, ctx, *, arg: str):
...
+                try:
+                    valid_roles.append(
+                        self.bot.get_guild(self.bot.config.server_id).get_role(
+                            int(role)))
+                except AttributeError:
+                    pass
+
+            roles_name = [
+                "[{}] {}\n".format(i + 1, role.name)
+                for i, role in zip(range(len(valid_roles)), valid_roles)
```
davidlougheed, line 605:
```python
for i, role in enumerate(valid_roles)
```

davidlougheed (changes requested):

cogs/helpers.py
le-potate: Resolved in ed39449

cogs/helpers.py
le-potate: Resolved in ed39449

cogs/helpers.py
le-potate: Resolved in ed39449

cogs/helpers.py
le-potate: Resolved in ed39449

cogs/helpers.py
le-potate: Resolved in ed39449

cogs/helpers.py
le-potate: Resolved in ed39449

cogs/helpers.py
```diff
@@ -563,6 +564,94 @@ async def colour(self, ctx, *, arg: str):
...
+        A moderator can click the OK react on the message to give these roles back
+        """
+        conn = sqlite3.connect(self.bot.config.db_path)
+        c = conn.cursor()
+        fetched_roles = c.execute(
+            'SELECT Roles FROM PreviousRoles WHERE ID = ?',
+            (user.id, )).fetchone()
+        # the above returns a tuple with a string of IDs separated by spaces
+        if fetched_roles is not None:
+            roles_id = list(fetched_roles[0].split(" "))
```
davidlougheed, line 593: remove list cast

cogs/helpers.py
```diff
@@ -563,6 +564,94 @@ async def colour(self, ctx, *, arg: str):
...
+            valid_roles = []
+            for role in roles_id:
+                try:
+                    valid_roles.append(
+                        self.bot.get_guild(self.bot.config.server_id).get_role(
+                            int(role)))
+                except AttributeError:
```
davidlougheed, line 600: maybe add a comment explaining where this attributeerror occurs from

cogs/helpers.py
```diff
@@ -563,6 +564,94 @@ async def colour(self, ctx, *, arg: str):
...
+                    pass
+
+            roles_name = [
+                "[{}] {}\n".format(i + 1, role.name)
+                for i, role in enumerate(valid_roles)
```
davidlougheed, line 605: should have been better reviewed the first time, my bad:
```python
for i, role in enumerate(valid_roles, 1)
```
to avoid adding 1 each time

davidlougheed (changes requested):

cogs/helpers.py
le-potate: Resolved in e375496

cogs/helpers.py
le-potate: This was a mistake, resolved in e375496

cogs/helpers.py
le-potate: Resolved in e375496

davidlougheed (approved):

PR #322, opened 2020-09-05
Improve assignable roles and add ?create_role

davidlougheed:
hehe


le-potate (approved):
good changes, will make it easier to convert this to database eventually, especially by moving stuff that mods might want to change like categories and limits to globals which should probably become db later. Using ?iam and ?iamn makes the commands way more intuitive than having to write `add` after the command.

cogs/roles.py
```diff
@@ -17,129 +15,228 @@
...
+        "pronouns",
+        "fields",
+        "faculties",
+        "years",
+        "generics",
+    )
+
+    CATEGORY_LIMITS = {
+        "pronouns": None,
+        "fields": 2,
```
krubenok, line 45: Limit here should be 3 IMO. Arts is major + 2 minors.

davidlougheed: 5c27cd3eb5efe0b0e56c02415ac994957a49cfea

le-potate:
Related to #136

PR #325, opened 2020-09-06
Fixes and quality of life changes for role assign

le-potate:
Related to #136

Why this change was necessary
Makes the command more intuitive and nice to use, fixes certain cases where roles couldn't be added, or roles could be added when they shouldn't

## What this change does
-Instead of saying category/ies detect if there is many categories
-Fix bug where specific commands like `?field` etc. could be used to add roles from the `generics` category since next() defaulted to `generics`.
-This bug was also causing every wrong role entered to appear as "invalid role for category `generics`", even if it should say that it is invalid for the category of the command (i.e. if ?field is called and the role isnt found it is invalid for `fields`)
-Fix capitalization, which made it so that roles that weren't written in all lowercase, uppercase, titlecase couldn't be added if the user didn't use the right capitalization
-Change error message that said a role does not exist to say that it should exist but doesn't since I think it is the only way that error could happen
-Now sends a message if a user tries to add a role they already have
-Now sends a message when a role is replaced by another (for the case where the limit is one)
-When a user has too many roles in a category, the message now says the limit
-Sends a message when a user tries to remove a role they do not have
-Added dots at the end of all messages sent to the bot for consistency
-Added plural aliases for all the specific commands like `?field`. Added `generic` to `?iam` since users could try using that since all the other categories have commands. Added `program` and `major` to field since it is also refered by those names instead
-Fix bug where when adding too much roles in a category using `?iam` it would say too many roles in categories `fields`, `faculties`, `years`, etc. instead of the category of the role.

## How was this tested?
I tested all the different possibilities again

## Types of changes

## Checklist:


cogs/roles.py
```diff
@@ -69,6 +69,7 @@ async def toggle_role(self, ctx, transaction: RoleTransaction,
         Assigns a single role to a user with no checks from a category of roles
         """
 
+        fcategory = f"{'categories' if len(categories) > 1 else 'category'}"
```
davidlougheed, line 72: format-string isnt needed, could just be
```python
fcategory = "categories" if len(categories) > 1 else "category"

File: cogs/roles.py [hunk 2]
```diff
@@ -79,66 +80,78 @@ async def toggle_role(self, ctx, transaction: RoleTransaction,
...
-                                           for r in self.roles[c]}),
-            "generics")
+        try:
+            category = next((c for c in categories if requested_role.lower() in
+                             {r.lower()
+                              for r in self.roles[c]}))
+        except StopIteration:
+            await ctx.send(
+                f"Invalid role for {fcategory} `{', '.join(categories)}`.")
+            return
```
davidlougheed, line 95: ye

cogs/roles.py
```diff
@@ -79,66 +80,78 @@ async def toggle_role(self, ctx, transaction: RoleTransaction,
...
+
         roles = self.roles[category]
 
-        if requested_role.lower() not in {r.lower() for r in roles}:
-            await ctx.send(f"Invalid role in category `{category}`")
-            return
+        roles_lower = [r.lower() for r in roles]
+        requested_role = roles[roles_lower.index(requested_role.lower())]
```
davidlougheed, line 100: nice

davidlougheed (approved):

cogs/roles.py
le-potate: Resolved in 4837c90

le-potate:
I added clarifications that says a list of roles is returned if no argument is given in the help for the different self-assign commands

davidlougheed (approved):

PR #328, opened 2020-09-14
Add Log Hierarchy, New Mod Logs Webhook

le-potate:
Logs for the mods shouldn't be accessible in the log channel for dev

**What this change does**
-Instead of one Webhook there is now one Dev Webhook and one Mod Webhook, both optionally specified in `config.ini`.
-Info like who used `?mix` goes to the Mod Webhook (in cogs, use `self.bot.mod_logger.info("message")`). Logs useful for devs goes to the Dev Webhook (In cogs, use `self.bot.dev_logger.info("message")`
-All these logs still go in the log file

**More precisions**:
In `bot.py`:
-Add log hierarchy: Create new "sub"loggers `Canary.Dev` and `Canary.Mod`. The loglevel for `Canary.Dev` is the same than for `Canary`, that is the one specified in `config.ini`. The loglevel for `Canary.Mod` is info, since that is the level at which info for mods is logged.
-As it already was, the "main" `Canary` logger is handled by the file handler. (Everything logged to the "sub"loggers is also sent up to `Canary`, and thus to the log file.)
-The `_WebhookHandler` class is generalized to work with the two "sub"loggers.
-The two "sub"loggers are both handled by their respective webhook handler.
-In `config.ini` and `parser.ini`:
`LogWebhookID` and `LogWebhookToken` are replaced by `DevLogWebhookID`, `DevLogWebhookID`, `ModLogWebhookID` and `ModLogWebhookID`
In `memes.py`:
-`?mix` and `?bac` are now logged to the mod logger
In main.py:
Now sends the message that says output is logged to webhook if `dev_log_webhook_id` and `dev_log_webhook_token` have values
Other:
-Every mention of `bot(or self).logger` is changed to `bot(or self).dev_logger`. This is the logger that should now be used for logs that are useful for devs. Using `logger` only sends the log to the log file, not the dev webhook. There were mentions in `bot.py` and `main.py`
-Updated the README

## Types of changes

## Checklist:

### DIFF
--- a/Main.py
+++ b/Main.py
@@ -62,14 +62,15 @@ MARTY_RESPONSES = {
 
 @bot.event
 async def on_ready():
-    if bot.config.log_webhook_id and bot.config.log_webhook_token:
+    if bot.config.dev_log_webhook_id and bot.config.dev_log_webhook_token:
         webhook_string = " and to the log webhook"
     else:
         webhook_string = ""
     sys.stdout.write(f'Bot is ready, program output will be written to a '
                      f'log file{webhook_string}.\n')
     sys.stdout.flush()
-    bot.logger.info('Logged in as {} ({})'.format(bot.user.name, bot.user.id))
+    bot.dev_logger.info('Logged in as {} ({})'.format(bot.user.name,
+                                                      bot.user.id))
 
 
 @bot.command()
@@ -108,7 +109,7 @@ async def restart(ctx):
     """
     Restart the bot
     """
-    bot.logger.info('Bot restart')
+    bot.dev_logger.info('Bot restart')
     await ctx.send('https://streamable.com/dli1')
     python = sys.executable
     os.execl(python, python, *sys.argv)
@@ -120,7 +121,7 @@ async def sleep(ctx):
     """
     Shut down the bot
     """
-    bot.logger.info('Received sleep command. Shutting down bot')
+    bot.dev_logger.info('Received sleep command. Shutting down bot')
     await ctx.send('Bye')
     await bot.logout()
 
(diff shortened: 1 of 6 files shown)


Main.py
```diff
@@ -62,14 +62,15 @@
...
-    if bot.config.log_webhook_id and bot.config.log_webhook_token:
+    if bot.config.dev_log_webhook_id and bot.config.dev_log_webhook_token:
         webhook_string = " and to the log webhook"
     else:
         webhook_string = ""
     sys.stdout.write(f'Bot is ready, program output will be written to a '
                      f'log file{webhook_string}.\n')
     sys.stdout.flush()
-    bot.logger.info('Logged in as {} ({})'.format(bot.user.name, bot.user.id))
+    bot.dev_logger.info('Logged in as {} ({})'.format(bot.user.name,
```
davidlougheed, line 72: we can use f-strings now, thanks yapf!!!!! :)))))))))))

Main.py
```diff
@@ -164,15 +165,16 @@ async def backup(ctx):
...
 if __name__ == "__main__":
     for extension in startup:
         try:
             bot.load_extension(extension)
         except Exception as e:
-            bot.logger.warning('Failed to load extension {}\n{}: {}'.format(
-                extension,
-                type(e).__name__, e))
+            bot.dev_logger.warning(
+                'Failed to load extension {}\n{}: {}'.format(
```
davidlougheed, line 177: f-string

davidlougheed (approved):

Main.py
le-potate: Resolved in 380013c

Main.py
le-potate: Resolved in 380013c

davidlougheed (approved):

PR #331, opened 2020-09-14
Fix Dockerfile and revise Travis builds

davidlougheed:
## Description

* Adds apt dependencies to Dockerfile
* Uses Docker exit codes directly for testing startup


idodin (approved):
Lgtm - thanks for updating gitignore as well

PR #335, opened 2020-09-17
inchannel function

### LINKED ISSUE #224: See the list of users in a channel
### Feature Description:
Since the McGill server now has more than 1000 users, it is not possible to see a complete list of users in any channel anymore. 

It is important that only people in a certain channel can see the list of users in that channel. A way to do this would be a command `?inchannel` that sends the list of users in the channel itself. Another way would be a command `?inchannel #channel` that sends a DM to the person who called if they are in the channel.

clu2001:
!!! Resolves #224 !!!

inchannel function - used to see list of users in the current channel

## Motivation and Context
- Resolves #224
- Lots of users in McGill server but only a few members are actually active and verified

## How Has This Been Tested?
Tested in idoneam server

## Types of changes
- Created a new function

## Checklist:


davidlougheed:
based motivation

cogs/roles.py
```diff
@@ -141,6 +141,32 @@ async def inrole(self, ctx, *, queryRole):
                           editable_content=False)
             await pages.paginate()
 
+    @commands.command()
+    async def inchannel(self, ctx):
+        """Returns list of users in current channel"""
+        members = None
```
davidlougheed, line 147: this line isnt needed

cogs/roles.py
```diff
@@ -141,6 +141,32 @@ async def inrole(self, ctx, *, queryRole):
...
+        channel = ctx.message.channel
+        members = channel.members
+
+        if members is None:
+            return
+
+        channelUsers = list(map(lambda m: str(m) + "\n", members))
```
davidlougheed, line 154: use underscores (`channel_users`) for consistency

cogs/roles.py
```diff
@@ -141,6 +141,32 @@ async def inrole(self, ctx, *, queryRole):
...
+        header = f"List of users in {channel.mention} - {len(members)}"
+
+        # TODO remove for paginator take empty list for embed
+        if len(channelUsers) == 0:
```
davidlougheed, line 158:
```python
if not channel_users:
```
is cleaner

davidlougheed (changes requested):

cogs/roles.py
```diff
@@ -141,6 +141,32 @@ async def inrole(self, ctx, *, queryRole):
...
 
+    @commands.command()
+    async def inchannel(self, ctx):
+        """Returns list of users in current channel"""
+        members = None
+        channel = ctx.message.channel
+        members = channel.members
+
+        if members is None:
+            return
```
le-potate, line 152: I don't think this could happen, you would get an empty list if there was no member, which anyway, isn't possible since presumably at least the bot and one user that calls the command are there

cogs/roles.py
le-potate: this isn't possible anyway, see my other review but there's always at least the bot and the person calling the command

le-potate (changes requested):

le-potate:
Good first PR 🎉

cogs/roles.py
le-potate: Resolved in b20d0ac

cogs/roles.py
le-potate: Resolved in b20d0ac

cogs/roles.py
le-potate: Resolved in b20d0ac

cogs/roles.py
le-potate: Resolved in b20d0ac

cogs/roles.py
```diff
@@ -257,6 +257,23 @@ async def create_role(self, ctx, *, role: Optional[str] = None):
...
+    @commands.command()
+    async def inchannel(self, ctx):
+        """Returns list of users in current channel"""
+        channel = ctx.message.channel
+        members = channel.members
+
+        channel_users = list(map(lambda m: str(m) + "\n", members))
+        header = f"List of users in {channel.mention} - {len(members)}"
+
+        # TODO remove for paginator take empty list for embed
```
le-potate, line 269: This comment isn't needed anymore

cogs/roles.py
```diff
@@ -257,6 +257,23 @@ async def create_role(self, ctx, *, role: Optional[str] = None):
...
         await ctx.send("Role created successfully.")
 
+    @commands.command()
+    async def inchannel(self, ctx):
+        """Returns list of users in current channel"""
+        channel = ctx.message.channel
+        members = channel.members
+
+        channel_users = list(map(lambda m: str(m) + "\n", members))
+        header = f"List of users in {channel.mention} - {len(members)}"
```
le-potate, line 267: Since this is used into an embed title, "channel.mention" doesn't work there, instead you get `<#channel_id>`. Use `#{channel}` instead

le-potate (changes requested):

cogs/roles.py
le-potate: Resolved in fc6d147

cogs/roles.py
le-potate: Resolved in fc6d147

le-potate (approved):

davidlougheed (approved):

PR #338, opened 2020-09-26
Add mute/unmute (penalty) functionality 

davidlougheed:
Adds a mute function with the ability to cache/restore roles and apply a penalty role

## Description
`?mute` places a user in penalty and removes all their other roles, storing them in a db table
`?unmute` removes user from penalty and restores their roles

## Motivation and Context
penalty is annoying to do by hand, and role overrides can lead to issues - this automates away that problem

## How Has This Been Tested?
tested on idoneam server (still more to be done, WIP)

## Types of changes

## Checklist:


Martlet.schema
```diff
@@ -49,7 +49,15 @@ CREATE TABLE IF NOT EXISTS `BankTransactions` (
...
 	`ID`	      INTEGER UNIQUE,
-	`Roles`       TEXT
+	`Roles`       TEXT -- space-separated list of role IDs
+);
+
+-- Used for when users are placed in the Penalty role; stores old roles akin to
+-- the way PreviousRoles does it.
+CREATE TABLE IF NOT EXISTS `PenaltyUsers` (
+    `ID`	      INTEGER UNIQUE,
+	`Roles`       TEXT -- space-separated list of role IDs
```
le-potate, line 62: seems like theres a mix of tabs and spaces there, but it's not the only place like that in the file

README.md
```diff
@@ -52,6 +52,8 @@ You must set certain values in the `config.ini` file, in particular your Discor
     * `DownvoteEmoji`: The name of your downvote emoji.
     * `ModeratorRole`: The name of the role that your moderators have (for functions like DMing users).
     * `DeveloperRole`: The name of the role that your developers have (for functions like restarting the bot). This could be the same role than moderator.
+    * `PenaltyRole`: The name of the role used for placing users in a "penalty box" (which presumably removes some
```
le-potate, line 55: i think this should be optional. right now, if you dont have the penaltyrole, you get the exception mis-configured penalty role everytime someone leaves

cogs/roles.py
```diff
@@ -257,8 +264,238 @@ async def create_role(self, ctx, *, role: Optional[str] = None):
...
+        return penalty_role and next(
+            (r for r in user.roles if r == penalty_role), None) is None
+
+    def _is_in_penalty_table(self, user: discord.Member):
+        conn = sqlite3.connect(self.bot.config.db_path)
+        try:
+            c = conn.cursor()
+            penalty = c.execute("SELECT * FROM PenaltyUsers WHERE ID = ?",
+                                (user.id, )).fetchone()
+            return penalty is None
```
le-potate, line 325: doesnt this do the opposite of what it's supposed to. If penality is none, then is_in_penalty_table should return False, not True. Right now, every new user gets the penalty role !

cogs/roles.py
```diff
@@ -257,8 +264,238 @@ async def create_role(self, ctx, *, role: Optional[str] = None):
...
+
+        if not roles_id:
+            return
+
+        conn = sqlite3.connect(self.bot.config.db_path)
+        try:
+            c = conn.cursor()
+            # store roles as a string of IDs separated by spaces
+            t = (user.id, " ".join(str(e) for e in roles_id))
+            c.execute(f"REPLACE INTO {table} VALUES (?, ?)", t)
```
le-potate, line 285: If a user gets muted but has no roles, they wont get added to the penalty table

le-potate (changes requested):

cogs/roles.py
davidlougheed: 🐒

davidlougheed: 678a97a

cogs/roles.py
davidlougheed: 9dd2412

README.md
davidlougheed: 06e0fe4

Martlet.schema
davidlougheed: 560f8ee

le-potate (approved):
I am currently working on a new muting/appeal system that will work a bit differently, but still use a lot of this. This is working so it's fine if the bot ends up being updated before I'm finished with the new system, but I think I will be done this week.

davidlougheed:
i am a god of code

PR #360, opened 2020-11-23
Add a Github Actions YAPF Task

krubenok:
## Description

Create a new GitHub actions workflow to lint the project with YAPF. Test passes if no issues are found. 

## Motivation and Context

- Github Actions is generally agreed as the future of CI and testing on the project and is more transparent than the existing Travis Solution. 
- By running only yapf and not any other testing in this action, it's a very lightweight test that can be run on every commit. Future actions could be added with more thorough testing that run less frequently (i.e. only on PR). 

## How Has This Been Tested?

Tested in Github Actions. (like this very PR :) )

## Types of changes

## Checklist:


namemcguffin (approved):
seems good, no criticisms to be made here

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,14 @@
...
+on: [push]
+
+jobs:
+  formatting-check:
+    name: Formatting Check
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: run YAPF to test if python code is correctly formatted
+      uses: AlexanderMelde/yapf-action@master
```
jidicula, line 12: Might be [safer](https://julienrenaux.fr/2019/12/20/github-actions-security-risk/) to use the commit hash versioning rather than pointing to a branch, no?

Also, why use a 3rd-party action rather than just pip-installing YAPF within the workflow and running YAPF on the repo?

krubenok: I largely agree if this were anything more than a YAPF test, but given that it seems like a pretty low risk thing to do. 

_Eventually_ Dependabot is supposed to support action updates as well so we can reevaluate then.

davidlougheed (approved):
lgtm

PR #370, opened 2020-12-30
Add P-strings (Rich placeholder strings)

le-potate:
A "string with placeholders" is a simple string like "hi %user%, this is the channel %channel%", which contains placeholders, here %user% and %channel%. A p-string is composed of a "string with placeholders", and values for these placeholders. Printing a p-string will fill the string with these values. This utility will be mostly used in cases where messages have to be set
from the server. For a simple example, a moderator might want to set the goodbye message to be one of "bye `@someone`" or "ciao `@someone`". They would thus set the goodbye_message to "%[bye, ciao]% %user%" from the server.
Then, when a user leaves, `await ctx.send(PString(goodbye_message, user=ctx.author.mention))` would be used,
which will send "bye `@someone`" or "ciao `@someone`"

Another, more complex use is groups placeholders, which will be used as part of input/output patterns. For example, there could be a custom reaction that takes a prompt "I like %1%" and sends a response "I don't like %1%". Then, if someone writes "I like apples" in the server, "I don't like apples" would be sent. This is done by making a PStringEncodings object with a list of input "strings with placeholders" and a list of output "strings with placeholders" (see PStringEncodings docstring). Calling PStringEncodings.parser on the content of a message will then detect if the message corresponds to some input "strings with placeholders", choose a random one, and create a filled output p-string with its corresponding output "string with placeholders".

## Placeholders:
 - %user%: [Only used for output strings] should generally be used with something like ctx.author.mention but can be used with any string. 

- %channel% [Only used for output strings] should generally be used with something like ctx.channel but can be used with any string

- choice list (%[a,b]%): a random value will be chosen from the list (here, either a or b). The values themselves can be choice lists
(ex: %[hi, hello %[world, you]%]%) will print either "hi", "hello world" or "hello you". They can also contain other placeholders, for example a valid input string with placeholders ould be `%[%1% is my name yes, I am]% %1% `

- groups (placeholder: %1%, %2%, etc... up to %9%): for example, if there is a p-string with string "%1% likes %2%" and
 groups=["He", "apples"], printing the p-string will return "He likes apples"

## Types of changes

## Checklist:


le-potate:
bump

cogs/utils/p_strings.py
```diff
@@ -0,0 +1,286 @@
...
+        choice = f"(?:{'|'.join(choice_list)})"
+    # otherwise, pick a random element of the list
+    else:
+        choice = random.choice(choice_list)
+    # replace everything between the %[ and ]% by the choice that was chosen
+    # or the pattern, and return a call to the function itself since there
+    # might still be pairs of %[ and ]% remaining
+    choice_list_string = (choice_list_string[:last_start_pos] + choice +
+                          choice_list_string[first_end_pos_after + 2:])
+    return _convert_choice_list(choice_list_string, level=level + 1)
```
davidlougheed, line 94: is level ever used?

cogs/utils/p_strings.py
```diff
@@ -0,0 +1,286 @@
...
+            filled_string = re.sub("%user%", self.user, filled_string)
+        if self.channel:
+            filled_string = re.sub("%channel%", self.user, filled_string)
+        # for every group placeholder, replace them by their value
+        if self.groups:
+            for i, group in enumerate(self.groups, 1):
+                filled_string = re.sub(f"%{i}%", group, filled_string)
+        # convert the choice lists
+        filled_string = _convert_choice_list(filled_string)
+        return filled_string
```
davidlougheed, line 175: simplify to 
```python
return _convert_choice_list(filled_string)
```

davidlougheed (changes requested):
overall looks cool

cogs/utils/p_strings.py
le-potate: Resolved in a55d8b9

cogs/utils/p_strings.py
le-potate: Resolved in a55d8b9

cogs/utils/p_strings.py
```diff
@@ -0,0 +1,285 @@
...
+        self.groups = groups
+        self.additional_info = additional_info
+
+    def __str__(self):
+        filled_string = self.string
+        # replace the %user% and %channel% placeholders by their values
+        if self.user:
+            filled_string = re.sub("%user%", self.user, filled_string)
+        if self.channel:
+            filled_string = re.sub("%channel%", self.user, filled_string)
```
davidlougheed, line 168: shouldnt this be subbing for self.channel? shoulda caught that first time my bad

davidlougheed (changes requested):

cogs/utils/p_strings.py
davidlougheed: maybe theres a way to have a dict of "replacements" with keys being the stuff between the % and values being the replacements to make this systematic and extensible

namemcguffin: to expand on david's point, this would allow us to precache the most common regexes, such as the number ones and the user and channel ones

cogs/utils/p_strings.py
```diff
@@ -0,0 +1,285 @@
...
+        self.user = user
+        self.channel = channel
+        self.groups = groups
+        self.additional_info = additional_info
+
+    def __str__(self):
+        filled_string = self.string
+        # replace the %user% and %channel% placeholders by their values
+        if self.user:
+            filled_string = re.sub("%user%", self.user, filled_string)
```
namemcguffin, line 166: i think we should precompile the `%user%` and `%channel%` regexes, as well as the %1% to %9% regexes

cogs/utils/p_strings.py
le-potate: Resolved in 8d3f102

cogs/utils/p_strings.py
le-potate: Resolved in 8d3f102

cogs/utils/p_strings.py
```diff
@@ -159,17 +162,18 @@ def __init__(self,
         self.groups = groups
         self.additional_info = additional_info
 
+    @property
+    def patterns_and_values(self):
```
davidlougheed, line 166: i'd make this private (add a _ in front)

davidlougheed (changes requested):

cogs/utils/p_strings.py
le-potate: Resolved in f780970

davidlougheed (approved):

PR #371, opened 2020-12-30
Update discord.py and add Intents

le-potate:
See https://discordpy.readthedocs.io/en/latest/intents.html

We now need to explicitely specify Intents when starting the bot. For now, `discord.Intents.default()` is every intents except intent.members and intent.presences, which is why they are manually set to True

## How Has This Been Tested?
Functions like `on_member_update` that need privileged intents are still working

## Types of changes

## Checklist:


namemcguffin:
tested it, seems to work fine

namemcguffin (approved):
yep seems good

davidlougheed:
lgtm

PR #374, opened 2020-12-31
Replace Travis Status by the new action check

le-potate:
Replace the old build statuses in the README by the new build statuses using the github action

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,9 +5,9 @@ Canary is a Python3 bot designed for the McGill University Community Discord Ser
 
 ## Build Statuses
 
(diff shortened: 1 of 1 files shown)


krubenok:
I guess this is accurate but like... I don’t think it’s really reflective of anything since it’s just YAPF 🤷‍♂️. 

Seems to me like a lot of things are being held up by not having any kind of functional tests (thorough or otherwise).

krubenok (approved):

le-potate:
Yeah agreed, although that's what the old Travis Build status reflected too since its the only thing it checked. You're right that eventually we need more useful checks, and with this at least there will be a nice template for when those actions come

PR #375, opened 2021-01-04
Keydates hotfix

namemcguffin:
Changes to keydates command for future-proofing, as well as some refractoring

## Description
divs with `notes` class are now deleted, as they are considered unnecessary
minor refractoring changes also made

## Motivation and Context
fix a bug and future proof command

## How Has This Been Tested?
tested in idoneam server

## Types of changes

## Checklist:


cogs/helpers.py
```diff
@@ -331,7 +331,12 @@ async def keydates(self, ctx):
                            colour=0xDA291C)
 
         for i in range(len(headers)):
-            em.add_field(name=headers[i], value=sections[i], inline=False)
+            if len(headers[i]) <= 256:
+                em.add_field(name=headers[i], value=sections[i], inline=False)
+            else:
+                em.add_field(name=f"{headers[i][:255]}\u2026",
```
davidlougheed, line 337:
```python
em.add_field(
    name=headers[i][:255] + "\u2026" if len(headers[i]) > 256 else "",
    value=sections[i],
    inline=False
)
```
and remove if statement outside

davidlougheed (changes requested):

davidlougheed (dismissed):
we should probably eventually make utils for trimming fields since this if/else could be used on all embed fields

davidlougheed (approved):

PR #380, opened 2021-01-08
Python version bump and dependency stabilization

namemcguffin:
bumps python version to 3.9, cleans up `Dockerfile` and stabilizes dependecies

## Description
made changes to `pyproject.toml` to bump python version as well as stabilize wildcarded packages
made changes to `Dockerfile` to bump python version, which also required adding a `gcc` dependency
also removed the poetry install in `Dockerfile`, since it isn't used

## Motivation and Context
allow use of walrus operator and new dict operators, as well as prepare for the 3.10 bump, which would be cool
so that we can use the match operator that's being added in that version

## How Has This Been Tested?
tested in idoneam discord

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.


le-potate:
dockerimage?

davidlougheed (approved):

PR #381, opened 2021-01-08
Miscellaneous refactoring and small fixes

### LINKED ISSUE #341: [Enhancement] If no semicolons present, use commas for choose
If I accidentally type `?choose a, b`  instead of `a; b` then it'll not give me what I expect

Check if `;` is present; if so split on that, otherwise split on `,`

### LINKED ISSUE #342: [Enhancement] Update assignable roles
Add `Continuing Studies` as a faculty, `Accounting` as a field (for div)

davidlougheed:
## Description

* Move LaTeX preamble for ?tex to outside the command, to avoid re-defining it every call and make it easier to update
* Moves @le-potate 's feels-like weather calculation into an enclosed function
* Added f-strings in place of `.format()` where useful/nice
* Fixed wording on TEPID status to avoid saying "at least one printer is down" when both are down
* Narrowed exception clauses in some places, or made notes of where it should be done
* Shrank history look-back for image commands to improve performance
* Make image processing functions properly static (because they are)
* Fix ?bac and ?mix errors on blank arguments
* Refactor `auto_incorrect` to reduce pointless lines / move defines out of function
* Refactor ?pyramid and limit the unit size to avoid obnoxious messages / API size limits
* Add some missing english equivalents in reminders and code-ify the replacements
* Remove some useless comments that restated the obvious
* Other misc. code styling stuff - snake case, useless encoding headers, copyright updates
* Closes #341 
* Closes #342 

## Motivation and Context
Trying to bring all our code up to modern Python standards, while fixing various tiny issues encountered over the past little while.

## How Has This Been Tested?
Commands have been tested on my own server.

## Types of changes

## Checklist:


cogs/helpers.py
```diff
@@ -89,66 +89,68 @@ async def exam(self, ctx):
...
-        Data taken from http://weather.gc.ca/city/pages/qc-147_metric_e.html"""
+        """
+        Retrieves current weather conditions.
+        Data taken from http://weather.gc.ca/city/pages/qc-147_metric_e.html
+        """
         def retrieve_string(label):
             return soup.find(
                 "dt", string=label).find_next_sibling().get_text().strip()
 
+        def calculate_feels_like(temp: float, humidity: float, ws_kph: float) \
```
namemcguffin, line 100: shouldn't this be defined outside the weather command? maybe either as a private static method, or as a function in utils?

cogs/helpers.py
```diff
@@ -89,66 +89,68 @@ async def exam(self, ctx):
 
     @commands.command()
     async def weather(self, ctx):
-        """Retrieves current weather conditions.
-        Data taken from http://weather.gc.ca/city/pages/qc-147_metric_e.html"""
+        """
+        Retrieves current weather conditions.
+        Data taken from http://weather.gc.ca/city/pages/qc-147_metric_e.html
+        """
         def retrieve_string(label):
```
namemcguffin, line 96: shouldn't this be defined outside the weather command? maybe either as a private static method, or as a function in utils?

cogs/helpers.py
davidlougheed: sure, private static just above is probably fine

cogs/helpers.py
davidlougheed: less sure about this one cause its tiny but yeah maybe

namemcguffin (approved):

namemcguffin (approved):
finally us undergrads can rise up and have the best role

namemcguffin (approved):

PR #387, opened 2021-01-09
[feat][CI] Check Poetry dependencies

jidicula:
We should check the Poetry dependency installation when changes are
made to `pyproject.toml`, `poetry.lock`, or the workflow files themselves.

**What this change does**
Creates workflows that run on PR and Push

**Any side-effects?**
We'll see

### DIFF
--- a/.github/workflows/poetry-dev.yml
+++ b/.github/workflows/poetry-dev.yml
@@ -30,7 +30,7 @@ jobs:
           - 'ubuntu-18.04'
           - 'ubuntu-16.04'
           - 'macos-10.15'
-          - 'macos-11.0'
+          # - 'macos-11.0'
           - 'windows-2019'
           - 'windows-2016'
         python-version:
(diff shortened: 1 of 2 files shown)


davidlougheed (changes requested):
If possible and to reduce the time / matrix of jobs, here's what i'd suggest:

 * Eliminate Ubuntu 16.04 because it's effectively EOL
 * Only run on windows-2019
 * Only run prod checks on PR to master, since issues where this breaks but dev doesnt should be rare and we'll then catch them before release anyway

jidicula:
> If possible and to reduce the time / matrix of jobs, here's what i'd suggest:
> 
> * Eliminate Ubuntu 16.04 because it's effectively EOL
> * Only run on windows-2019
> * Only run prod checks on PR to master, since issues where this breaks but dev doesnt should be rare and we'll then catch them before release anyway

Resolved in 86e3798

jidicula:
Note that `poetry-prod` no longer gets triggered in this PR since it's not a PR to `master`.

davidlougheed (approved):

PR #406, opened 2021-03-06
feat: add music functions

namemcguffin:
Adds a `Music` cog to the bot, allowing users to play audio from youtube

## Description
+ added music `Music` cog, which contains the following functions, with a FIFO song queue:
  + `music|m`: monocommand for music functions
+ adds `yt_dlp` as a pip dependency
+ adds `ffmpeg` as a dependency to the `Dockerfile`

## Missing Features / Comments
this PR does not (and will not) include any DB integrated playlist features, even though such features are standard for music bots
this is was done for two reasons

+ it would add a significant amount of code to what is already a very large PR, thus making playlist features a separate PR allows this PR to remain (relatively) focused in its scope
+ there are currently plans to rework the DB system for the bot, as such making a PR right now that utilizes the current DB structure would most likely mean either more work for the DB restructuring PR or a second PR after the DB restructuring is done to fix the DB integration here

## Motivation and Context
reduce reliance on mosebot

## How Has This Been Tested?
tested extensively in the idoneam server

## Types of changes

## Checklist:


Main.py
```diff
@@ -44,7 +44,8 @@
     "cogs.roles",
     "cogs.score",
     "cogs.subscribers",
-    "cogs.customreactions",    # Do not remove this terminating comma.
+    "cogs.customreactions",
+    "cogs.music",    # Do not remove this terminating comma
```
davidlougheed, line 48: it'd be nice to get this back into alphabetical order

cogs/music.py
```diff
@@ -0,0 +1,246 @@
...
+        if "entries" in data:
+            if len(data["entries"]) > 1:
+                self.song_queue.extend(data["entries"])
+                await ctx.send(f"queued up playlist: `{data.get('title')}`")
+            else:
+                data = data["entries"][0]
+                self.song_queue.append(data)
+                await ctx.send(f"queued up track: `{data.get('title')}`")
+        else:
+            self.song_queue.append(data)
```
davidlougheed, line 190: deduplicate this with the above else, maybe with something like
```py
data = await self.get_info(url)

entries = data.get("entries", [data])
self.song_queue.extend(entries)

if len(entries) > 1:
    await ctx.send(f"queued up playlist: `{data.get('title')}`")
    return

await ctx.send(f"queued up track: `{entries[0].get('title')}`")
```

davidlougheed (changes requested):

davidlougheed (changes requested):
use early return + factor out repetitive checks into a decorator for commands stop, skip, pause, and resume

cogs/music.py
```diff
@@ -0,0 +1,584 @@
...
+
+        def release_lock(_):
+            self.song_lock.release()
+
+        if url:
+            await ctx.trigger_typing()
+            data = await self.get_info(url)
+            entries = data.get("entries", [data])
+            for track in reversed(entries):
+                self.song_queue.insert(0, (track, str(ctx.author)))
```
davidlougheed, line 181: deque's `extendleft` method might be able to make this a "one-liner" with an iterator - plus, entries is only used in one place here

cogs/music.py
```diff
@@ -0,0 +1,584 @@
...
+            if not in_main:
+                ctx.voice_client.stop()
+
+        if in_main:
```
davidlougheed, line 185:
```python
if not in_main:
    return

#...
```

cogs/music.py
```diff
@@ -0,0 +1,584 @@
...
+                now_playing = None
+                if self.skip_opts is None:
+                    await ctx.trigger_typing()
+                    if ctx.voice_client is None:
+                        break
+                    if (not self.song_queue) and (self.looping is None):
+                        if self.loop_queue:
+                            self.song_queue = self.backup
+                            self.backup = deque()
+                        else:
```
davidlougheed, line 199: early break here
```python
if not self.loop_queue:
    break
# ...
```

cogs/music.py
```diff
@@ -0,0 +1,584 @@
...
+            await ctx.author.voice.channel.connect()
+
+            while True:
+                await self.song_lock.acquire()
+                now_playing = None
+                if self.skip_opts is None:
+                    await ctx.trigger_typing()
+                    if ctx.voice_client is None:
+                        break
+                    if (not self.song_queue) and (self.looping is None):
```
davidlougheed, line 195: don't need these parens

cogs/music.py
```diff
@@ -0,0 +1,584 @@
...
+                value=mk_title_string(self.playing[0]),
+                inline=False,
+            ).add_field(
+                name="volume", value=f"{self.volume_level}%",
+                inline=True).add_field(
+                    name="duration",
+                    value=mk_duration_string(self.playing[0]),
+                    inline=True).add_field(
+                        name="looping",
+                        value="no" if self.looping is None else "yes",
```
davidlougheed, line 387: I would write this as 
```python
"yes" if self.looping else "no"
```

cogs/music.py
```diff
@@ -0,0 +1,584 @@
...
+            return
+        entries = data.get("entries", [data])
+        if not entries:
+            await ctx.send("could not find track")
+            return
+        author_str = str(ctx.author)
+        for track in reversed(entries):
+            self.song_queue.insert(song_index, (track, author_str))
+        if len(entries) > 1:
+            inserted = discord.Embed(
```
davidlougheed, line 441: duplicate code between here and the else branch (with the inserted thing, which could be `{'playlist' if len(entries) > 1 else 'track'}`)

you could define this outside the branch, then add the fields as needed conditionally

cogs/music.py
```diff
@@ -0,0 +1,584 @@
...
+                colour=random.randint(0, 0xFFFFFF),
+                title=f"inserted playlist at index {song_index}").add_field(
+                    name="playlist name",
+                    value=mk_title_string(data),
+                    inline=False).add_field(name="duration",
+                                            value=mk_duration_string(data),
```
davidlougheed, line 447: data is the same as `entries[0]`, so there might be deduplication that can be done with the playlist dialog here

cogs/music.py
```diff
@@ -0,0 +1,584 @@
...
+        except youtube_dl.utils.DownloadError:
+            await ctx.send("could not find track")
+            return
+        entries = data.get("entries", [data])
+        if not entries:
+            await ctx.send("could not find track")
+            return
+        author_str = str(ctx.author)
+        for track in entries:
+            self.song_queue.append((track, author_str))
```
davidlougheed, line 486: use `extend` instead of for loop + append

cogs/music.py
```diff
@@ -0,0 +1,584 @@
...
+        if len(entries) > 1:
+            queued = discord.Embed(colour=random.randint(0, 0xFFFFFF),
+                                   title="queued up playlist").add_field(
+                                       name="playlist name",
+                                       value=mk_title_string(data),
```
davidlougheed, line 491: data = entries[0], so this could maybe be deduplicated with below

davidlougheed (changes requested):

cogs/music.py
```diff
@@ -0,0 +1,553 @@
...
+        removed = mk_change_embed(track_list[del_index][0],
+                                  [track_list[del_index][0]],
+                                  f"removed track at index {track_index}",
+                                  f"removed by: {ctx.author}")
+        del track_list[del_index]
+        await ctx.send(embed=removed)
+
+    @commands.command(aliases=["iqs", "iqt"])
+    @check_banned
+    async def insert_track(self, ctx, track_index: int, *, url: str):
```
le-potate, line 402: right now this doesnt work if theres no song or just one song but i think it should

cogs/music.py
```diff
@@ -0,0 +1,553 @@
...
+        for _ in range(queue_amount % (len_q + 1)):
+            self.backup.append(self.track_queue.popleft())
+        ctx.voice_client.stop()
+        await ctx.send(
+            f"skipped current track{f' and {queue_amount} more from the queue' if queue_amount else ''}."
+        )
+
+    @commands.command(aliases=["prev", "previous", "rev"])
+    @check_playing
+    async def rewind(self, ctx, queue_amount: int = 1):
```
le-potate, line 501: this should be called `back` or `previous`

cogs/music.py
```diff
@@ -0,0 +1,553 @@
...
+    async def track_loop(self, ctx):
+        """Changes track looping state"""
+
+        self.looping_track = not self.looping_track
+        await ctx.send(
+            f"current track will now{' ' if self.looping_track else ' no longer '}loop."
+        )
+
+    @commands.command(aliases=["ql"])
+    async def queue_loop(self, ctx):
```
le-potate, line 270: i'd rename this `loop_queue`

cogs/music.py
```diff
@@ -0,0 +1,553 @@
...
+    @check_playing
+    @time_func
+    def backward_time(self, seconds: int, paused: bool):
+        """Move backwards in currently playing track"""
+
+        return max(0, round(self.compute_curr_time(paused) - seconds))
+
+    @commands.command(aliases=["tl", "song_loop", "sl"])
+    @check_playing
+    async def track_loop(self, ctx):
```
le-potate, line 261: i'd rename this loop_track

le-potate (changes requested):
I think we need to rethink how this command is called. It would probably be a good idea to move everything under one command and do something like `?music play x` `?music pause` etc, but maybe an embed utility could be good too for some of these functions. For now I've just pointed out some argument names that I think will need to be changed regardless

Right now it is not consistent what people can or can't do if they're outside the voice channel. For example, someone outside of VC can pause the music that people are listening to, but is blocked from resuming it.

le-potate:
Bot also seems to stays in VC when music has ended

cogs/music.py
```diff
@@ -0,0 +1,686 @@
...
+        """
+        changes looping state
+        arguments: (either `track` or `queue`)
+        """
+
+        match target:
+            case "queue" | "q":
+                self.looping_queue = not self.looping_queue
+                await ctx.send(f"queue will now{' ' if self.looping_queue else ' no longer '}loop.")
+            case "track" | "t":
```
davidlougheed, line 388: add `song`

cogs/music.py
```diff
@@ -0,0 +1,686 @@
...
+        - help
+        """
+
+        if subcommand is None:
+            return await ctx.send(self.music.help)
+        if subcommand == "help":
+            if args is None:
+                return await ctx.send(self.music.help)
+            try:
+                tup = self.subc_decision(args)
```
davidlougheed, line 197: do you ever use tup[1]? if not, consider something like `fn, _ = self.subc...`

davidlougheed (changes requested):

cogs/utils/music_helpers.py
```diff
@@ -0,0 +1,116 @@
...
+            return
+        seconds = func(self, parsed, ctx.voice_client.is_paused())
+        self.skip_opts = time.strftime("%H:%M:%S",
+                                       time.gmtime(seconds)), seconds
+        self.track_start_time = time.perf_counter() - seconds
+        ctx.voice_client.stop()
+        await ctx.send(
+            f"moved to `{self.skip_opts[0]}` in currently playing track.")
+
+    wrapper.__name__ = func.__name__
```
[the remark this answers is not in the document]
namemcguffin: there is, its functools.wraps, but i don't use it here because it also overrides the function the wrapper type signature with the input function (which i don't wanna do here because i'm changing the output signature on purpose)

cogs/music.py
```diff
@@ -0,0 +1,663 @@
...
+        self.speed_flag = "atempo=1"
+
+    @check_playing
+    async def playback_speed(self, ctx, speed: float):
+
+        self.speed_val = max(0.25, min(speed, 4.0))
+        self.speed_flag = (
+            f"atempo={self.speed_val}"
+            if 0.5 < self.speed_val < 2
+            else f"atempo=sqrt({self.speed_val}),atempo=sqrt({self.speed_val})"
```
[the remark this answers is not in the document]
namemcguffin: ffmpeg only allows changing speed between 0.5 and 2 at once, so to circumvent that u can tell ffmpeg to do it twice

cogs/music.py
namemcguffin: i don't but i wanna isolate source of error here as much as possible so i think adding the extra unpacking isn't good

not super set on this tho so if you disagree i'll change it

cogs/music.py
```diff
@@ -0,0 +1,699 @@
...
+            case "loop":
+                return (self.loop, conv_arg(lambda x: x, True))
+            case "print_queue" | "pq":
+                return (self.print_queue, conv_arg(lambda x: 0 if x is None else int(x), False))
+            case "status" | "now_playing" | "np":
+                return (self.music_status, None)
+            case "remove" | "pop":
+                return (self.remove_track, conv_arg(int, True))
+            case "insert":
+                return (self.insert_track, conv_arg(lambda x: (lambda y, z: (int(y), z))(*x.split(maxsplit=1)), True))
```
davidlougheed, line 144: define a named function with some comments to explain this one rather than a one-liner as originally discussed

davidlougheed (changes requested):
after this, lgtm

davidlougheed (approved):

lazho (approved):
lgtm

PR #413, opened 2021-04-01
fix various issues with image functions

namemcguffin:
fixes various issues with our current image functions

## Description
- changed type signatures in image commands that take `str` arguments that internally get converted to `int` to `int` to use `discord.py` converters
- `opencv`'s blocking functions are now run using a `ProcessPoolExecutor` to avoid blocking the bot, as the functions can be CPU intensive
- images larger than 8MB now get compressed to 8MB to avoid using too much RAM which can lead to crashing the bot if it is being run in a container with limited RAM availability 

## Motivation and Context
fix issues described in the changes above

## How Has This Been Tested?
tested in the idoneam server, ready for review

## Types of changes

## Checklist:


cogs/images.py
```diff
@@ -20,45 +20,80 @@
...
+        else:
+            await ctx.send("image format not supported.")
+            return
+
         fn = f"{original_name}-{func.__name__}.{ext}"
+        ratio = (MAX_IMAGE_SIZE /
+                 att.size) * 100 if att.size > MAX_IMAGE_SIZE else 100
 
         try:
-            img_bytes = np.asarray(bytearray(buffer.read()), dtype=np.uint8)
```
trungams, line 50: maybe better to throw this whole chunk of code into a single lambda and give it to the executor instead of translating line by line

cogs/images.py
```diff
@@ -20,45 +20,80 @@
...
-            result = cv2.imdecode(img_bytes, cv2.IMREAD_UNCHANGED)
-            args = (args[0], result) if len(args) == 2 else (result, )
-            result = await func(self, ctx, *args)
-            _r, buffer = cv2.imencode(f".{ext}", result,
-                                      [cv2.IMWRITE_JPEG_QUALITY, 100])
+
+            with concurrent.futures.ProcessPoolExecutor() as pool:
```
trungams, line 63: I'd prefer thread pool instead of process pool, plus you should initialize this resource at bot startup (assign a bot.cpu_thread_pool or something). We should not depend on the function to implement that

cogs/images.py
```diff
@@ -20,45 +20,80 @@
 from discord.ext import commands
 
 # misc imports
+import concurrent.futures
 import os
 import numpy as np
 import cv2
 import math
-from functools import wraps
+from functools import wraps, partial
```
trungams, line 28: while you're at it, maybe add a decorator for the cpu intensive tasks :wink:

trungams (changes requested):
maybe we should support modpow on a image/2d matrix :thinking:

cogs/images.py
namemcguffin: python documentation states that `ProcessPoolExecutor`s should be used for CPU bound tasks (see: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.run_in_executor)
i don't think that having a `ProcessPoolExecutor` running the entire time is worth it either, considering the overhead, because the image functions aren't used that often, whereas the cost of starting a pool isnt that high anyways, especially considering the opencv2 functions are what take up the majority of the time/resources in these commands

cogs/images.py
namemcguffin: lambdas can't be passed through `run_in_executor` due to a pickling issue i don't fully understand

cogs/images.py
namemcguffin: after doing some more testing, it seems like the overhead on keeping a `ProcessPoolExecutor` in the background really is minimal, so ill change that

cogs/images.py
namemcguffin: refactored to do this

cogs/images.py
namemcguffin: due to various issues with how python passes functions to `run_in_executor` and pickling, i don't think this is currently possible

cogs/images.py
```diff
@@ -16,234 +16,97 @@
...
 
     @commands.command()
-    @filter_image
-    async def cart(self, _ctx, image=None):
+    async def cart(self, ctx):
         """
         Transform from polar to Cartesian coordinates.
         """
-        return Images._cart(np.rot90(image))
+        await ih.fitler_image(self.bot.loop, ih.cart, ctx, self.hist_lim,
```
davidlougheed, line 52: why is it fitler

cogs/utils/image_helpers.py
```diff
@@ -0,0 +1,194 @@
...
+
+    await ctx.trigger_typing()
+    buffer = BytesIO()
+    await att.save(fp=buffer)
+
+    original_name, ext = att.filename.rsplit(".", 1)
+    ext = ext.lower()
+    if ext in ("jpeg", "jpg"):
+        is_png = False
+    elif ext in ("png"):
```
davidlougheed, line 47: this is not correct, since `("png")` is not a tuple - should be `== "png"` or `in ("png",)` but i think the former is better

davidlougheed (changes requested):

davidlougheed (approved):

davidlougheed (approved):

PR #414, opened 2021-04-02
move sensitive keys to `.env` file

namemcguffin:
moves oauth keys from `config.ini` to `.env`

## Description
- added `.env.example`
- added `.vscode` subdirectory to `.gitignore` (incidental)
- changed `parser.py` to use `.env` keys instead of reading from `config.ini`
- removed sensitive key placeholder from `config.ini`
- added `python-dotenv` dependency (parses `.env` file) 

## Motivation and Context
`config.ini` currently holds both various sensitive keys as well as bot configuration settings
this means that some values that would like to be conserved and public are mixed with sensitive information, leading to issues when trying to switch branches, as well as making it quite easy to accidentally push these keys here, leading to a data breach

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:


krubenok:
I don't really see the point to this TBH. It was fine in Config.ini

PR #354, opened 2020-11-13
add hangman game

namemcguffin:
Added hangman game

## Description
added hangman game that can be invoked with either `?hm` or `?hangman` for a random category word
or with `?hm x` or `?hangman x` where` x` is one of the valid categories
also has a help message, callable with `?hm help` or `?hangman help`

additions:

+ `hangman.py` in `./cogs/utils/`, which is used to generate the hangman word lists, and which contains the `HangmanState` class used in the hangman command
+ `hangman_dict.obj` in `./data/premade/` directory, which contains word lists for the hangman game
+ `hangman` command to `games.py`
+ `HANGMAN_REWARD` action to `currency.py` 

changes:

+ renamed `requests.py` in `./cogs/utils/` to `custom_requests.py` so that it doesn't conflict when importing `requests` (which is needed in `hangman.py`)
+ changed imports in `helpers.py` (due to above change)

## Motivation and Context
ported missing functionality from Mose II 

## How Has This Been Tested?
tested extensively in the idoneam discord

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

PR #426, opened 2021-05-30
Add hangman game

namemcguffin:
see previously done PR


le-potate (approved):
This is the same PR as #354 which was previously reviewed and approved. That PR was reverted to fix history (it wasn't squashed, which we are now doing: settings have now been changed to enforce this). No commit had been made since that PR was merged

PR #427, opened 2021-05-30
Fix is_moderator and is_developer checks

le-potate:
Made this 7 months ago and forgot to make a PR apparently 😄 

-Fixed the lines were the MissingPermission exception is raised: was supposed to take a list of roles as input. 
-is_developer was sending the missing perm for is_moderator


le-potate:
Will remake this from most recent of dev for simplicity

PR #464, opened 2021-08-23
fix: Query metric units for wttr command

jidicula:
## Description

See title. See https://github.com/chubin/wttr.in#weather-units for API documentation.

## Motivation and Context

wttr command outputs a PNG with US units instead of metric. Likely caused by hosting from the US, since wttr.in selects units based on IP location of the query, and that this used to not be an issue when we were self-hosted.

## How Has This Been Tested?

Verified URL with browser and curl.

## Screenshots (if appropriate):

Before:
image

After: 
image

## Types of changes

## Checklist:


le-potate (approved):
lgtm

PR #564, opened 2022-08-29
New muting/appeal channels system

le-potate:
New muting system. Users get muted either by using the mute command or by being given the muted role. Once this happens, all other roles are saved and removed, and an appeal channel for that user is created. When unmuted (either using unmute command or removing roles), the previous roles can be given back, and the appeal channel is deleted. I've tried to make it so that this doesn't break completely if the channel is deleted manually, the role is removed while the bot isn't on, etc.

The part that saves roles when a user is muted was mostly written by @davidlougheed

## How Has This Been Tested?

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

### DIFF
--- a/Martlet.schema
+++ b/Martlet.schema
@@ -82,19 +82,19 @@ CREATE TABLE IF NOT EXISTS `PreviousRoles` (
 -- and stores old roles akin to the way PreviousRoles does it
 CREATE TABLE IF NOT EXISTS `MutedUsers` (
     `UserID`              INTEGER UNIQUE,
-	`AppealChannelID`     INTEGER UNIQUE,
+    `AppealChannelID`     INTEGER UNIQUE,
     `Roles`               TEXT, -- space-separated list of role IDs
-	`Date`                INTEGER
+    `Date`                INTEGER
 );
 
 CREATE TABLE IF NOT EXISTS `Settings` (
-	`Key`       TEXT PRIMARY KEY,
-	`Value`     TEXT
+    `Key`       TEXT PRIMARY KEY,
+    `Value`     TEXT
 );
 
 CREATE TABLE IF NOT EXISTS `BannerSubmissions` (
-	`UserID`              INTEGER UNIQUE,
-	`PreviewMessageID`    INTEGER,
-	`ConvertedMessageID`  INTEGER
+    `UserID`              INTEGER UNIQUE,
+    `PreviewMessageID`    INTEGER,
+    `ConvertedMessageID`  INTEGER
 );
 
--- a/Martlet.schema
+++ b/Martlet.schema
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) idoneam (2016-2019)
+ * Copyright (C) idoneam (2016-2022)
  *
  * This file is part of Canary
  *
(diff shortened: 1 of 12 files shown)


cogs/utils/paginator.py
```diff
@@ -93,11 +97,20 @@ def __init__(
...
+        if ctx:
+            self.bot = ctx.bot
+            self.guild = ctx.guild
+            self.channel = ctx.channel
+            self.user = ctx.author
+        else:
+            self.bot = bot
+            self.guild = guild
+            self.channel = channel
+            self.user = user
```
davidlougheed: personally, I'd rewrite this to allow ctx to either be the discordpy context or a custom context class which mimics these properties, which can be passed in as an alternative
so ctx would have the signature of roughly
```py
from typing import Union
# ...
ctx: Union[discord.Context, MockContext]
```

cogs/mod.py
```diff
@@ -111,6 +141,222 @@ async def finalize_crabbo(self, ctx):
...
+        if message.channel in self.muted_users_to_appeal_channels.values():
+            muted_user = self.muted_users_to_appeal_channels.inverse[message.channel]
+            if message.author == muted_user:
+                # only pings if the user is the one that is muted
+                log_message = f"Muted user {message.author.mention} ({message.author}) deleted message {message.id} in {message.channel.mention}"
+            else:
+                log_message = f"User {message.author} deleted message {message.id} in the appeal channel for muted user {muted_user.mention} ({muted_user}), {message.channel.mention}"
+            await self.appeals_log_channel.send(log_message)
+
+    # the next two functions are used to save the roles
```
davidlougheed: i only see one function

Martlet.schema
```diff
@@ -78,11 +78,13 @@ CREATE TABLE IF NOT EXISTS `PreviousRoles` (
...
--- Used for when users are placed in the Penalty role; stores old roles akin to
--- the way PreviousRoles does it.
-CREATE TABLE IF NOT EXISTS `PenaltyUsers` (
-    `ID`          INTEGER UNIQUE,
-    `Roles`       TEXT -- space-separated list of role IDs
+-- Used for when users are muted; stores the ID of the Appeal Channel created for that user
+-- and stores old roles akin to the way PreviousRoles does it
+CREATE TABLE IF NOT EXISTS `MutedUsers` (
+    `UserID`              INTEGER UNIQUE,
+	`AppealChannelID`     INTEGER UNIQUE,
```
davidlougheed: whats up with the indents here

cogs/roles.py
```diff
@@ -248,246 +246,40 @@ async def create_role(self, ctx, *, role: Optional[str] = None):
...
-        await pages.paginate()
+    @commands.Cog.listener()
+    async def on_member_remove(self, user: discord.Member):
+        if is_in_muted_table(self, user):
+            # Check if the user is muted.
+            # If so, save all roles BUT the muted role into the PreviousRoles table
+            save_existing_roles(self, user, muted=True)
+        else:
+            # Save existing roles
+            save_existing_roles(self, user)
```
davidlougheed: can this maybe be factored into
```py
@commands.Cog.listener()
async def on_member_remove(self, user: discord.Member):
    save_existing_roles(self, user, muted=is_in_muted_table(self, user))
```

also passing self as an explicit param is sort of non-conventional; you could make a mixin class and use multiple inheritance, or pass in the DB path or the bot object instead perhaps

davidlougheed (dismissed):

cogs/utils/paginator.py
le-potate: resolved in 8af4976

Martlet.schema
le-potate: resolved in 8af4976

cogs/roles.py
le-potate: resolved in 8af4976

cogs/mod.py
le-potate: resolved in 8af4976

namemcguffin (approved):

PR #565, opened 2022-08-29
add periodic purge of pictures in verif channel

le-potate:
This deletes all pictures from the verification channel once a week

## How Has This Been Tested?

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

### DIFF
--- a/cogs/mod.py
+++ b/cogs/mod.py
@@ -19,14 +19,68 @@ import discord
 import sqlite3
 import random
 from discord import utils
-from discord.ext import commands
+from discord.ext import commands, tasks
 
 from .utils.checks import is_moderator
+from datetime import datetime, timedelta
 
 
 class Mod(commands.Cog):
     def __init__(self, bot):
         self.bot = bot
+        self.guild = None
+        self.verification_channel = None
+        self.last_verification_purge_datetime = None
+
+    @commands.Cog.listener()
+    async def on_ready(self):
+        self.guild = self.bot.get_guild(self.bot.config.server_id)
+        self.verification_channel = utils.get(self.guild.text_channels, name=self.bot.config.verification_channel)
+
+        # arbitrary min date because choosing dates that predate discord will cause an httpexception
+        # when fetching message history after that date later on
+        self.last_verification_purge_datetime = datetime(2018, 1, 1)
+        conn = sqlite3.connect(self.bot.config.db_path)
+        c = conn.cursor()
+        c.execute("SELECT Value FROM Settings WHERE Key = ?", ("last_verification_purge_timestamp",))
+        fetched = c.fetchone()
+        if fetched:
+            last_purge_timestamp = float(fetched[0])
+            if last_purge_timestamp:
+                self.last_verification_purge_datetime = datetime.fromtimestamp(last_purge_timestamp)
+        else:
+            # the verification purge info setting has not been added to db yet
+            c.execute(
+                "INSERT INTO Settings VALUES (?, ?)",
+                ("last_verification_purge_timestamp", self.last_verification_purge_datetime.timestamp()),
+            )
+            conn.commit()
+        conn.close()
+
+        self.check_verification_purge.start()
+
+    @tasks.loop(minutes=60)
+    async def check_verification_purge(self):
+        # todo: make general scheduled events db instead
+        if not all((self.guild, self.verification_channel, self.last_verification_purge_datetime)):
+            return
+
+        if datetime.now() < self.last_verification_purge_datetime + timedelta(days=7):
+            return
+
+        conn = sqlite3.connect(self.bot.config.db_path)
+        c = conn.cursor()
+        # delete everything since the day of the last purge, including that day itself
+        await self.verification_purge_utility(self.last_verification_purge_datetime - timedelta(days=1))
+        # update info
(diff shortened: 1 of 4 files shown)


cogs/mod.py
```diff
@@ -19,14 +19,68 @@
...
         self.bot = bot
+        self.guild = None
+        self.verification_channel = None
+        self.last_verification_purge_datetime = None
+
+    @commands.Cog.listener()
+    async def on_ready(self):
+        self.guild = self.bot.get_guild(self.bot.config.server_id)
+        self.verification_channel = utils.get(self.guild.text_channels, name=self.bot.config.verification_channel)
+
```
davidlougheed, line 52: if verification channel isn't set, maybe this should exit early / not do anything, to make it an optional feature before even touching the db

cogs/mod.py
```diff
@@ -111,6 +165,48 @@ async def finalize_crabbo(self, ctx):
 
         await ctx.message.delete()
 
+    async def verification_purge_utility(self, after):
+        # after can be None, a datetime or a discord message
```
davidlougheed, line 201: use type hints here to avoid a comment maybe

`<py3.10`:
```py
from typing import Union
# ...
async def verification_purge_utility(self, after: Union[datetime.datetime, discord.Message, None]):
    pass
```
`>=py3.10`:
```py
async def verification_purge_utility(self, after: datetime.datetime | discord.Message | None):
    pass
```

something like that

cogs/mod.py
```diff
@@ -19,14 +19,68 @@
...
 from .utils.checks import is_moderator
+from datetime import datetime, timedelta
 
 
 class Mod(commands.Cog):
     def __init__(self, bot):
         self.bot = bot
+        self.guild = None
+        self.verification_channel = None
+        self.last_verification_purge_datetime = None
```
davidlougheed: type hints

cogs/mod.py
```diff
@@ -111,6 +165,48 @@ async def finalize_crabbo(self, ctx):
...
 
+    async def verification_purge_utility(self, after):
+        # after can be None, a datetime or a discord message
+        await self.verification_channel.send("Starting verification purge")
+        channel = self.verification_channel
+        deleted = 0
+        async for message in channel.history(oldest_first=True, limit=None, after=after):
+            if len(message.attachments) > 0 or len(message.embeds) > 0:
+                content = message.content
+                if len(message.embeds) > 0:
```
davidlougheed:
```py
if message.embeds:
```

cogs/mod.py
```diff
@@ -111,6 +165,48 @@ async def finalize_crabbo(self, ctx):
 
         await ctx.message.delete()
 
+    async def verification_purge_utility(self, after):
+        # after can be None, a datetime or a discord message
+        await self.verification_channel.send("Starting verification purge")
+        channel = self.verification_channel
+        deleted = 0
+        async for message in channel.history(oldest_first=True, limit=None, after=after):
+            if len(message.attachments) > 0 or len(message.embeds) > 0:
```
davidlougheed:
```py
if message.attachments or message.embeds:
```

cogs/mod.py
```diff
@@ -111,6 +165,48 @@ async def finalize_crabbo(self, ctx):
...
+    @is_moderator()
+    async def verification_purge(self, ctx, id: int = None):
+        """ "
+        Manually start the purge of pictures in the verification channel.
+
+        If a message ID is provided, every pictures after that message will be removed.
+        If no message ID is provided, this will be done for the whole channel (may take time).
+        """
+        message = None
+        if id:
```
davidlougheed: i'd prefer `id is not None` here since id can technically also be 0 although this would never happen here, but as a pattern for ints explicitly checking None is better practice

davidlougheed (changes requested):

cogs/mod.py
le-potate: Resolved in 02632a4

cogs/mod.py
le-potate: Resolved in 02632a4

cogs/mod.py
le-potate: Resolved in 02632a4

cogs/mod.py
le-potate: Resolved in 02632a4

cogs/mod.py
le-potate: Resolved in 02632a4

cogs/mod.py
le-potate: Resolved in 684c87e

cogs/mod.py
```diff
@@ -36,28 +36,28 @@ def __init__(self, bot):
     async def on_ready(self):
         self.guild = self.bot.get_guild(self.bot.config.server_id)
         self.verification_channel = utils.get(self.guild.text_channels, name=self.bot.config.verification_channel)
+        if self.verification_channel:
```
davidlougheed: it'd be better to do an early return here

davidlougheed (dismissed):

cogs/mod.py
le-potate: resolved in e6c43e4

davidlougheed (approved):

PR #566, opened 2022-08-29
Spoilerize

le-potate:
New feature that lets mods repost messages as spoilers. The messages appear as though they have been sent by the user itself. Users can also use this feature for themselves, although it is not that useful in that case.

This also has an atomic commit of arg_converter which now has two default converters.

## How Has This Been Tested?

## Types of changes

## Checklist:

### DIFF
--- a/cogs/helpers.py
+++ b/cogs/helpers.py
@@ -1,4 +1,4 @@
-# Copyright (C) idoneam (2016-2021)
+# Copyright (C) idoneam (2016-2022)
 #
 # This file is part of Canary
 #
@@ -40,7 +40,10 @@ import random
 from .utils.paginator import Pages
 from .utils.custom_requests import fetch
 from .utils.site_save import site_save
+from .utils.checks import is_moderator, is_developer
 import sqlite3
+from .utils.arg_converter import ArgConverter, StrConverter
+from discord.ext.commands import MessageConverter
 
 MCGILL_EXAM_URL = "https://www.mcgill.ca/exams/dates"
 
@@ -64,10 +67,17 @@ LATEX_PREAMBLE = r"""\documentclass[varwidth,12pt]{standalone}
 \usepackage{amsmath,amsfonts,lmodern}
 \begin{document}"""
 
+MAIN_WEBHOOKS_PREFIX = "Main webhook for #"
+
 
 class Helpers(commands.Cog):
     def __init__(self, bot):
         self.bot = bot
+        self.guild = None
+
+    @commands.Cog.listener()
+    async def on_ready(self):
+        self.guild = self.bot.get_guild(self.bot.config.server_id)
 
     @commands.command(aliases=["exams"])
     async def exam(self, ctx):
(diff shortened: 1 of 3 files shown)


cogs/helpers.py
```diff
@@ -64,10 +67,17 @@
 \usepackage{amsmath,amsfonts,lmodern}
 \begin{document}"""
 
+MAIN_WEBHOOKS_PREFIX = "Main webhook for #"
+
 
 class Helpers(commands.Cog):
     def __init__(self, bot):
         self.bot = bot
+        self.guild = None
```
davidlougheed: type hint it (whatever guild class | None)

cogs/helpers.py
```diff
@@ -639,6 +649,209 @@ async def translate(self, ctx, command: str, *, inp_str: str = None):
...
+                    # max length for webhook name is 80
+                    await channel.create_webhook(
+                        name=f"{MAIN_WEBHOOKS_PREFIX}{channel.name[:(80 - len(MAIN_WEBHOOKS_PREFIX))]}"
+                    )
+                    await ctx.send(f"Created webhook for {channel.mention}")
+                else:
+                    await ctx.send(f"Webhook for {channel.mention} already exists")
+            except discord.errors.Forbidden:
+                ignored += 1
+                continue
```
davidlougheed: needless continue

cogs/helpers.py
```diff
@@ -639,6 +649,209 @@ async def translate(self, ctx, command: str, *, inp_str: str = None):
...
+        Create the general-use webhooks for each channel. Ignores channels that already have them
+        """
+        ignored = 0
+        for channel in self.guild.text_channels:
+            try:
+                webhook = discord.utils.find(
+                    lambda w: w.user == self.bot.user and w.name[: len(MAIN_WEBHOOKS_PREFIX)] == MAIN_WEBHOOKS_PREFIX,
+                    await channel.webhooks(),
+                )
+                if not webhook:
```
davidlougheed: id prefer a `if webhook: ctx.send, continue` type pattern here - then you can outdent the main creation stuff

cogs/helpers.py
```diff
@@ -639,6 +649,209 @@ async def translate(self, ctx, command: str, *, inp_str: str = None):
...
+            c.execute("REPLACE INTO SpoilerizedMessages VALUES (?, ?)", (spoilerized_message.id, author.id))
+        conn.commit()
+        conn.close()
+
+        # delete original message
+        await message.delete()
+
+    @commands.Cog.listener()
+    async def on_raw_reaction_add(self, payload):
+        if payload.guild_id == self.guild.id and str(payload.emoji) == "🚮":
```
davidlougheed: maybe inverse condition and early return, then add a comment saying what the main body below does

cogs/helpers.py
```diff
@@ -639,6 +649,209 @@ async def translate(self, ctx, command: str, *, inp_str: str = None):
...
+        except commands.BadArgument as err:
+            raise commands.BadArgument(str(err))
+        msg_1 = args_dict["id_1"]
+        msg_2 = args_dict["id_2"]
+        reason = " ".join([args_dict[f"reason_{i}"] for i in range(len(args)) if args_dict[f"reason_{i}"]])
+
+        if msg_1 and ctx.message.reference and ctx.message.reference.resolved:
+            await ctx.send("Cannot specify a message ID to spoilerize and use the command as a reply at the same time")
+            return
+        elif msg_1 and not msg_2:
```
davidlougheed: maybe add line break and change this to `if`, since above if returns

davidlougheed (changes requested):

cogs/helpers.py
le-potate: Resolved in 913212b

cogs/helpers.py
le-potate: Resolved in 913212b

cogs/helpers.py
le-potate: Resolved in 913212b

cogs/helpers.py
le-potate: Resolved in 913212b

cogs/helpers.py
le-potate: Resolved in 913212b

davidlougheed (approved):

